PR #617, opened 2018-11-28
Fixes PHP Parse error:  syntax error, unexpected ')'

CurtisBelt:
Since it's in the dev folder, this file probably doesn't actually get accessed -- I found this when running a script to preload opcache, which compiles every php file on my system. But hey I thought I'd send in a fix anyway! :smile:


herrvigg:
Yes good catch, it is legacy code for the slugs that i simply moved and did not try yet but for the least it should compile. Thanks!

PR #624, opened 2018-12-10
Activates the gravity forms module

reddo:
The module wasn't being activated.


herrvigg:
You got it right! My bad, it was certainly due to the last rebase, i made it a bit quickly :-/

I have also to merge this with woo-commerce so it will be easier to manage, coming soon.

PR #627, opened 2018-12-11
Modules for integrated plugins (WooCommerce, ...)

herrvigg:
New feature!
Support for integration modules in qTranslate-XT to replace the legacy integration plugins:
- woo-commerce -> legacy plugin [WooCommerce & qTranslate-X](https://github.com/qTranslate-Team/woocommerce-qtranslate-x)
- gravity-forms -> legacy plugin [qTranslate support for GravityForms](https://github.com/wp-plugins/qtranslate-support-for-gravityforms)
- wp-seo -> legacy plugin [Integration: Yoast SEO & qTranslate-X](https://github.com/qTranslate-Team/wp-seo-qtranslate-x) (aka WP SEO)

This new module engine is generic:
- checks are first performed on the admin side, detecting the candidate modules from active plugins. Incompatible legacy plugins are deactivated if possible. The active modules are then stored in the new `qtranslate_modules` option.
- the available modules are then loaded both on the client and/or admin side depending on the module itself, but the list is retrieved from the `qtranslate_modules` option so the global overhead should be very limited.

New integration modules can be added later.


herrvigg:
I have now added a a new module called wp-seo for Yoast SEO from the legacy plugin [Integration: Yoast SEO & qTranslate-X](https://github.com/qTranslate-Team/wp-seo-qtranslate-x). Note its slug is wordpress-seo and all its variables are `wpseo_xxx` so i kept this as module slug.

It was a bit confusing to understand how it works, because there was a `i18n-config.json` in this repo, but as i understand it, it's not used. All the config (both admin and front) is actually set by the code. Maybe the intention was to move to the json file, but nevermind, i kept this json file aside for reference.

Now we need some testers before we can merge this. Even there are some small bugs, we should try to do it soon so we can move forward!

One feature i'm thinking about is to list the modules in the admin panel on the options page and maybe have some options there to enable/disable every module separately. Let's see if this is feasible without too much hassle.

PR #630, opened 2018-12-15
Xabispacebiker show alternative content prefix

xabispacebiker:
This patch, gives the option to choose whether or not to display the "For the sake of viewer convenience" text when alternative content is enabled.

### DIFF
--- a/admin/qtx_configuration.php
+++ b/admin/qtx_configuration.php
@@ -385,6 +385,13 @@ function qtranxf_conf() {
                                                                              value="1"<?php checked( $q_config['show_alternative_content'] ) ?>/> <?php _e( 'Show content in an alternative language when translation is not available for the selected language.', 'qtranslate' ) ?>
                                 </label>
                                 <p class="qtranxs-notes"><?php printf( __( 'When a page or a post with an untranslated content is viewed, a message with a list of other available languages is displayed, in which languages are ordered as defined by option "%s". If this option is on, then the content of the first available language will also be shown, instead of the expected language, for the sake of user convenience.', 'qtranslate' ), __( 'Default Language / Order', 'qtranslate' ) ) ?></p>
+				<br />
+                                <label for="show_alternative_content_prefix"><input type="checkbox"
+                                                                             name="show_alternative_content_prefix"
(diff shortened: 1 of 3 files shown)


herrvigg:
hello, thank you for this PR! I'll check that soon :)

herrvigg:
It's a good feature, but we should regroup the checkbox options into a list for better clarity, also because all combinations don't work anyway. If i find a bit of time i'll give a try.

herrvigg:
@xabispacebiker i'm thinking to convert the `show_alternative_content` option from boolean to enum (int), the retro-compatibility should be straightforward if the mapping is done correctly.
Could you rebase on master (or just merge the last commits)? Then i'll try to push to your branch after testing this (there is a github option to allow this).

herrvigg:
Actually i don't like this "For the sake of viewer convenience" notice at all. It's quite cumbersome. We might even just remove it for simplicity or maybe have this in a tooltip instead. Let's see.

herrvigg:
@xabispacebiker this took a bit of time, but i have finally reviewed all this in deep.

In fact my suggestion is to **never display** the very long message "_For the sake of convenience [long text...]_" because it's cumbersome and it has **no added value for the user**! Any user can understand what an hyperlink is and there's no real need explaining why a given language is currently displayed. The goal of your option was to hide this but i want now to make it the new behavior instead. Therefore i will not take your PR but i will transform your idea a bit differently.

Actually the message "_For the sake..._" is more of a debug information for the administrator, not the user! But the admin should know what he is doing when the option "**Show alternative content**" is selected and how this alternative content is built, first from the default and then from the order of the other languages (this is actually one of the best features of qTranslate that i really love!).

As a consequence i'm simply removing all this block so this also simplifies the code. The only complication comes from the filter `i18n_content_translation_not_available` that some people might use. I hardly see someone wanting to re-use this message so i just want to get rid of it. But the content might still be customized entirely by the filter. As i see it the best solution is to deprecate the filter to remove this argument. I will try this and i think i can send it soon.

herrvigg:
I replaced with PR https://github.com/qtranslate/qtranslate-xt/pull/655. Let me know what you think of it.

PR #633, opened 2019-01-13
Update qtx_import_export.php

alphp:
"continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in E:\var\www\html\wp\wp-content\plugins\qtranslate-xt\admin\qtx_import_export.php on line 50

### DIFF
--- a/admin/qtx_import_export.php
+++ b/admin/qtx_import_export.php
@@ -47,7 +47,7 @@ function qtranxf_migrate_options_copy( $nm_to, $nm_from ) {
 			case 'qtranslate_page_configs':
 			case 'qtranslate_admin_config':
 			case 'qtranslate_front_config':
-				continue;
+				continue 2;
 			default:
 				break;
 		}


herrvigg:
Are you really sure it is meant to be `continue 2` and not `break`?

It's the first time i ever see something related to the migration. But the current instruction should be equivalent to `break` before PHP7.3, as stated in this warning and in the related [RFC](https://wiki.php.net/rfc/continue_on_switch_deprecation). Also, what makes me doubt is this comment:
> "It is now easier to list options which need to be copied, instead".

However, if we consider the current `continue` to be equivalent `break` the whole `switch` loop is completely pointless. It doesn't make any sense!! :-/

alphp:
http://php.net/manual/en/control-structures.continue.php
> **Note:** In PHP the switch statement is considered a looping structure for the purposes of continue. continue behaves like break (when no arguments are passed). If a switch is inside a loop, continue 2 will continue with the next iteration of the outer loop.

I have simplified the foreach loop. I think that now we avoid jumps that produce mistakes and the thread of execution is clearer.

```php
	// $skip_options => Array with any option to skip
	$skip_options[] = 'qtranslate_flag_location';
	$skip_options[] = 'qtranslate_admin_notices';
	$skip_options[] = 'qtranslate_domains';
	$skip_options[] = 'qtranslate_editor_mode';
	$skip_options[] = 'qtranslate_custom_fields';
	$skip_options[] = 'qtranslate_custom_field_classes';
	$skip_options[] = 'qtranslate_text_field_filters';
	$skip_options[] = 'qtranslate_qtrans_compatibility';
	$skip_options[] = 'qtranslate_header_css_on';
	$skip_options[] = 'qtranslate_header_css';
	$skip_options[] = 'qtranslate_filter_options_mode';
	$skip_options[] = 'qtranslate_filter_options';
	$skip_options[] = 'qtranslate_highlight_mode';
	$skip_options[] = 'qtranslate_highlight_mode_custom_css';
	$skip_options[] = 'qtranslate_lsb_style';
	$skip_options[] = 'qtranslate_custom_i18n_config';
	$skip_options[] = 'qtranslate_config_files';
	$skip_options[] = 'qtranslate_page_configs';
	$skip_options[] = 'qtranslate_admin_config';
	$skip_options[] = 'qtranslate_front_config';

	// search in $options array options to be copied
	foreach ( $options as $option ) {
		$name = $option->option_name;
		//skip new specific options
		if (!in_array($name, $skip_options) and !strpos($name, '_flag_location')) {
			$value = maybe_unserialize( $option->option_value );
			$nm = str_replace( $nm_from, $nm_to, $name );
			update_option( $nm, $value );
		}
	}
```

herrvigg:
Semantically speaking, the solution with `continue 2` looked fine. It is equivalent and even better to me. However, it's an unexpected change in the behavior of the code.

In PHP before 7.3, a `continue` in a switch under a loop was equivalent to a `break`. Seen there is no code either after the option case or the default case, the whole `switch` block is completely useless and could have been entirely removed. The current code does not skip these options. Do you agree with that?

Now if the correct solution is a `continue 2` to skip these options, it would mean it's been bugged for a very long time without anyone to notice it! 

I don't feel sure enough to change this right now. I'd like other people to check at this and validate this function better before.

alphp:
https://wiki.php.net/rfc/continue_on_switch_deprecation
http://php.net/manual/en/migration73.incompatible.php#migration73.incompatible.core.continue-targeting-switch

The behavior isn't changed.
Now, in PHP-7.3.x, "continue" in "switch" statement throw a warning.

The loop has always had that bug, but PHP has never warned it.

It looks better with "continue 2", but it reads and keeps better in the way I propose.

I could have used an [array_diff](http://php.net/manual/en/function.array-diff.php), as I usually do in my programs. He could also have used another way of declaring the array with the values to satar. But it was about improving the readability of the code.

alphp:
Personally I prefer to avoid jumps of more than one level, in general, I avoid the use of "continue" whenever I can.

Another way to declare the array that maintains readability:
```php
	$skip_options = [
		'qtranslate_flag_location',
		'qtranslate_admin_notices',
		'qtranslate_domains',
		'qtranslate_editor_mode',
		'qtranslate_custom_fields',
		'qtranslate_custom_field_classes',
		'qtranslate_text_field_filters',
		'qtranslate_qtrans_compatibility',
		'qtranslate_header_css_on',
		'qtranslate_header_css',
		'qtranslate_filter_options_mode',
		'qtranslate_filter_options',
		'qtranslate_highlight_mode',
		'qtranslate_highlight_mode_custom_css',
		'qtranslate_lsb_style',
		'qtranslate_custom_i18n_config',
		'qtranslate_config_files',
		'qtranslate_page_configs',
		'qtranslate_admin_config',
		'qtranslate_front_config',
	];
```

herrvigg:
> The behavior isn't changed.
> Now, in PHP-7.3.x, "continue" in "switch" statement throw a warning.
> The loop has always had that bug, but PHP has never warned it.

I mean the behavior is changed with this PR. There was apparently a bug that no one raised and your change is fixing it.

I tend to believe it should be fixed as you say (any of the two solutions gives the same result), but i'm still quite surprised it only comes now through this PHP warning... so i'm a bit confused with this.

But if you can please double-check and ensure me this is how it should work i can merge this.

alphp:
I have created a quick test replicating the loop and its functionality to test it with PHP 5.6, 7.2 and 7.3:

```php
<?php
	$numbers = array_keys(array_fill(0, 10, null));

	echo '$numbers array:', PHP_EOL;
	echo implode("\t", $numbers), PHP_EOL;

	echo PHP_EOL;
	echo 'Original loop:', PHP_EOL;
	foreach ($numbers as $number) {
		switch ($number) {
			case 1:
			case 3:
			case 5:
			case 7:
			case 9:
				continue;
			default:
				break;
		}
		echo $number, "\t";
	}

	echo PHP_EOL;
	echo 'continue 2 loop:', PHP_EOL;
	foreach ($numbers as $number) {
		switch ($number) {
			case 1:
			case 3:
			case 5:
			case 7:
			case 9:
				continue 2;
			default:
				break;
		}
		echo $number, "\t";
	}

	$skip_numbers = [1, 3, 5, 7, 9];

	echo PHP_EOL;
	echo '$skip_numbers loop:', PHP_EOL;
	foreach ($numbers as $number) {
		if (!in_array($number, $skip_numbers)) {
			echo $number, "\t";
		}
	}
```

Result:
```
E:\usr\test>..\php-5.6.40\php continue.php
$numbers array:
0       1       2       3       4       5       6       7       8       9

Original loop:
0       1       2       3       4       5       6       7       8       9
continue 2 loop:
0       2       4       6       8
$skip_numbers loop:
0       2       4       6       8

E:\usr\test>..\php-7.2.12\php continue.php
$numbers array:
0       1       2       3       4       5       6       7       8       9

Original loop:
0       1       2       3       4       5       6       7       8       9
continue 2 loop:
0       2       4       6       8
$skip_numbers loop:
0       2       4       6       8

E:\usr\test>..\php-7.3.1\php continue.php

Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in E:\usr\test\continue.php on line 16
$numbers array:
0       1       2       3       4       5       6       7       8       9

Original loop:
0       1       2       3       4       5       6       7       8       9
continue 2 loop:
0       2       4       6       8
$skip_numbers loop:
0       2       4       6       8

```

I have the proposed PR in an installation in production and in another test, in both works without problems.

herrvigg:
all right! :)
It's also what i expected but do you agree with me the current code before the PR is like the `original loop`? Your PR not only removes the warning but it also changes the loop like the last two versions you show here (either `continue 2` or `skip_numbers`). This is what i mean by "changing the behavior".

What is really surprising me **a lot** is that the `original loop` has been like this for a very long time! This is the earliest  version i can track in this master here from 2015: 

https://github.com/qtranslate/qtranslate-xt/blob/83b4a9b513e623df3e9800888c742683c51eed6a/admin/qtx_import_export.php

```
	foreach ($options as $option)
	{
		$name = $option->option_name;
		//skip new qTranslate-X specific options
		switch($name){
			case 'qtranslate_flag_location':
			case 'qtranslate_admin_notices':
			case 'qtranslate_domains':
			case 'qtranslate_editor_mode':
			case 'qtranslate_custom_fields':
			case 'qtranslate_custom_field_classes':
			case 'qtranslate_text_field_filters':
			case 'qtranslate_qtrans_compatibility':
			case 'qtranslate_header_css_on':
			case 'qtranslate_header_css':
			case 'qtranslate_filter_options_mode':
			case 'qtranslate_filter_options':
				continue;
			default: break;
		}
		//if(strpos($name,'_flag_location')>0) continue;
		$value = maybe_unserialize($option->option_value);
		if(strpos($name,'_flag_location')>0) continue;
		$nm = str_replace($nm_from,$nm_to,$name);
		update_option($nm,$value);
```

In this code, you could remove the whole switch, it would be the same! How is that even possible it was bugged like this and no one noticed it? Even worse: as you can see, there were less options at this time. So they have added new options **supposedly to be skipped**, for example `qtranslate_front_config`, but they **were not** beause of the same (buggy?) loop. It means no one has really checked or used this before?!

Sorry if i'm asking you so many times, but do you arrive to the same conclusion? Can you confirm all of these options are to be skipped and therefore there was a bug?

Now another thing, there was also this check: `if(strpos($name,'_flag_location')>0) continue; `
In your code you replaced with the opposite check `!strpos($name, '_flag_location')` which should be  correct (keep both `false` and `0`).
Therefore we could remove `qtranslate_flag_location` from the corrected skip options list. But this flag check looks strange. Do you know if there is something else like `{xxx}_flag_location` to skip? What other prefix can be found here?

Thanks!

herrvigg:
What you mean by "3. Delete the entire loop"? You still want to skip these options right?
How it is done is not important to me.
What i want to understand is the functionality that was intended to work and if it didn't, to be sure the fix is correct and will not create some new regression. Is there a bug in the current code (before your PR)?!

alphp:
> Is there a bug in the current code (before your PR) ?!

YES

Option 3, eliminate the loop, are actually two options:
1. Do not skip any option, as it has worked so far:
```php
	foreach ( $options as $option ) {
		$name = $option->option_name;
		//skip new specific options
		if (!strpos($name, '_flag_location')) {
			$value = maybe_unserialize( $option->option_value );
			$nm = str_replace( $nm_from, $nm_to, $name );
			update_option( $nm, $value );
		}
	}
```
2. Remove the loop.

Being conservatives we can choose options 1 and 2.
What I'm sure of is that this code only runs when the plugin is activated.

The code:
```php
if (!strpos($name, '_flag_location')) {}
```
is identical to:
```php
if (strpos($name, '_flag_location') > 0) {}
```

alphp:
I mean, the code is an error because the creator had a reason to skip those options and they were not jumping.
The first version that skips options with a switch is 3.1:
https://plugins.svn.wordpress.org/qtranslate-x/tags/3.1/admin/import_export.php
In 3.0 I just skipped the options with "_flag_location" in the name:
https://plugins.svn.wordpress.org/qtranslate-x/tags/3.0/admin/import_export.php
If we review the story we can see that there are versions with the "_flag_location" commented:
https://plugins.svn.wordpress.org/qtranslate-x/tags/3.2/admin/import_export.php

All this is very confusing and it makes me wonder if this function "qtranxf_migrate_options_copy" is really necessary since it would only be useful to convert a really old installation with the old plugin.

alphp:
Perhaps the best thing is to correct the loop bug with a new code that works exactly the same as how that loop should have always worked.

Fix now and clean later or do it now?

alphp:
By the way, initially the plugin was the qTranslate, then qTranslate-X emerged as a continuity to the abandonment of qTranslate ... and now the story is repeated with name change to qTranslate-XT

herrvigg:
OK with the fix. Let's hope it's what was initially intended to be.

> By the way, initially the plugin was the qTranslate, then qTranslate-X emerged as a continuity to the abandonment of qTranslate ... and now the story is repeated with name change to qTranslate-XT

Well qTranslate-XT is only a transitory project because the author of qTranslate-X disappeared.
I didn't really mean to maintain this kind of specific code for import-export, import these functions should be taken separately imo, unlike the integration plugins. If we create a  new official plugin i'm not sure we would keep this. But for now it's fine.

Thanks :)

Issue #639, opened 2019-01-22
XSS vulnerability in IE (only IE)


arku31:
https://github.com/qtranslate/qtranslate-xt/blob/35ab2931a52f9d81ca4d016de7a1640ff7c7a14a/qtranslate_core.php#L1268

**How to check?**
Open in IE (not edge) http://live.paresh.ninja/redirect.php?u=https://localhost/en/"><svg/onload=alert(document.domain)>/%252e%252e/x

**How to fix**
Please, escape this with esc_url
`$complete = esc_url(qtranxf_buildURL($urlinfo_lang,$homeinfo));`

herrvigg:
hello @arku31 
thanks for this! A few questions:
- if that's only for IE isn't that more of a security issue of that browser?
- which version of IE? I tried with IE 11 (11.253.17763.0) but didn't see anything special. What is that suppose to do, show the alert message?
- which URL mode are you using? I suppose QTX_URL_PATH which is the default. Do you think it will work also with QTX_URL_DOMAIN / QTX_URL_DOMAINS / QTX_URL_QUERY? The domains modes are hard to test...!
- if the `urlinfo_lang` already contains something escaped, what happens? I'm not sure how risky the change is, i just want to be sure it won't create some regressions.

arku31:
- This is a security issue of that browser but that's IE... :) It's no longer supported but still in use
- Any IE. I forgot to mention, you should call this function, I'm using it like this `<?php echo qtrans_convertURL(get_permalink(), 'en'); ?>`
- Right, the default one
- Shouldn't change anything, but... I found regressions as well =( link with "&" are also encoded, so it brokes website, links with "&" leads to permanent redirect. I'm looking for another solution ATM.

herrvigg:
@arku31 After looking at this i find it a bit too specific and out of scope for qTranslate.

Since this is a security issue that modern browsers apparently deal with, it's more of a hack to compensate for a limitation in the old IE, because it's not maintained anymore. Though the proposed change is very small, and i thank you for that, i'd like to avoid adding an extra layer of complexity only for an outdated IE 😬 ... that no one should use anymore :sweat_smile:

What i suggest is you try to use the hook `qtranslate_url_set_language`. Ideally you'd like a hook after `qtranxf_buildURL` (there's actually some commented code about it!). But maybe by looking at the fields `$urlinfo [wp-path]` and `$urlinfo[query]` you could be able to handle it with the existing hook just before. Let me know if it works.

[...]

PR #640, opened 2019-01-22
fix IE xss #639

arku31:
Just escaping final URL to prevent XSS in IE


herrvigg:
I have looked at the code and we can't simply substitute `qtranxf_buildURL` with `qtranxf_sanitize_url`. They don't have the same role so this can create regressions. Look carefully at `qtranxf_buildURL`, there is much more being done. See also the questions in https://github.com/qtranslate/qtranslate-xt/issues/639. I close this until those are clarified.

PR #641, opened 2019-01-23
Fix AMP url for themes like Penci Soledad

tonyfahrion:
Hi,

I had an issue with the Penci Soledad AMP plugin, which caused qtranslate-xt to modify the localised URL twice (once before `/amp/` and once after). This resulted in a 404 for all internal links.

While I could not figure out the real cause (lack of time), this fixes
the localised URL string. It does so by looking
for a string like `/en/amp/en/` and replacing it as `/en/amp/`, so
both AMP and qtranslate-xt are happy and can be used in parallel.

This is tested with Penci Soledad AMP, but not with the official
AMP plugin from Wordpress/Google.

Please review

### DIFF
--- a/qtranslate_core.php
+++ b/qtranslate_core.php
@@ -1127,6 +1127,14 @@ function qtranxf_url_set_language( $urlinfo, $lang, $showLanguage ) {
 			case QTX_URL_PATH: // pre-path
 				//qtranxf_dbg_log_if(!isset($urlinfo['wp-path']),'qtranxf_url_set_language: wp-path not set $urlinfo:',$urlinfo,true);
 				$urlinfo['wp-path'] = '/' . $lang . $urlinfo['wp-path'];
+
+				// fix potential bug with AMP sites which leads to broken links like `/en/amp/en/...
+				// do this by replacing `/en/amp/en/` by `/en/amp/`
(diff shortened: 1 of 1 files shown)


herrvigg:
Thanks for raising this. I didn't look in the details yet, but i'm wondering if this shouldn't rather be handled by hooks such as `qtranslate_url_set_language` ? This is a specific integration problem and qTranslate-XT should not deal with all possible cases. At first sight the hook looks more suitable for this.

herrvigg:
@tonyfahrion after checking this again i'd like to avoid too specific features, the code is already over-bloated. The best is to use the existing filter: `qtranslate_url_set_language` on your side. This is called after setting the initial `$urlinfo['wp-path']`. So it should be enough to write a callback in your plugin/theme and you stick your code in there. If don't know how to, let me know.

tonyfahrion:
thx for the information @herrvigg 

Sorry, I don't have the time right now to investigate any further. But I'll come back to this in a few weeks 🙈 
But I have to agree, it might be a too specific problem.

Kind regards, and thank you to keep going with this wonderful plugin! 🙂 
Tony

herrvigg:
No worry i'm sure you'll figure out how to do, but if you get stuck don't hesitate to come back here. Good luck :)

PR #647, opened 2019-03-04
Add composer support

cheplv:
Add composer.json to support composer installations, currently added release under packagist repo "cheplv/qtranslate-x" with tag 3.5.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "cheplv/qtranslate-x",
-    "version": "3.5.3",
+    "version": "3.5.4",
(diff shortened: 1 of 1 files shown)


herrvigg:
hello,
i understand the need. But there's the question of the repo and the maintainer, it should be the same as for this repo with a stable reference, otherwise it's going to create a lot of confusion. It's already quite a mess. We first need to solve the question of the name and reference to qTranslate before anything. So i prefer to close it for now, sorry about that, but be sure we'll come back to it.

cheplv:
No problem - anyway i'll added support in my packagist repo - please ping me when this ticket actualized

PR #651, opened 2019-03-10
Using strtotime() instead of time() + $seconds

alphp:
Use of the strtotime function instead of time.
The strtotime function avoids errors by DST.


herrvigg:
Are you sure it's enough to use `strtotime`? Anyway it shouldn't be a big issue for the gettext DB but `time()` is used in many other places in the code so we should be looked at this potential problem everywhere. Everything looks quite minor because it's only about notices and so on. But if we need such a fix we should be consistent all over the whole project.

alphp:
Yes, in [my blog I talked](https://lab.fawno.com/2015/11/03/calculos-con-fechas-en-php/) some time ago about date calculations and the use of strtotime instead of time.

I've reviewed the use of time and I think I made all the necessary changes.

Some examples: https://lab.fawno.com/code/dates.php

herrvigg:
OK thanks, looks good!
Sorry i forced push a small commit amend to master just to revert a wrong permission change from 644 to 755 that i pushed by mistake. Can you rebase on master?

PR #653, opened 2019-03-11
Using strtotime() instead of time() + $seconds

alphp:
Use of the strtotime function instead of time.
The strtotime function avoids errors by DST.


alphp:
I got confused with git and in the end I ended up creating the branch with the changes.

herrvigg:
Yeah sometimes it's easier on small branches when the rebase goes banana ;)
Sorry for this mess but my WSL seems to have gone nuts with the file permissions only recently. Sad because it was a really convenient setup.
Thanks for the PR!

PR #655, opened 2019-03-12
Shorten message for alt content

herrvigg:
This PR is in substitution of https://github.com/qtranslate/qtranslate-xt/pull/630. This concerns the option "Show alternative content", meant to display the _post content in alternative languages_ when there's no translation for the current language. But instead of adding a new option to remove a part of the message, this becomes the new behavior.

Currently when the "alt content" option is selected many things are shown:
- notice: "Sorry, this entry is only available in %LANG: ...", containing links to allow language switch.
- long message: "For the sake of convenience, the content is shown below [very long text...]"
- the post content in alternative language (default or next in order)

With this PR the goal is to **remove** entirely the second part ("For the sake of convenience...").
As explained in https://github.com/qtranslate/qtranslate-xt/pull/630 it's cumbersome and it has no added value for the user! The first part should be sufficient, it's self-explanatory. Any user can understand what an hyperlink is and there's no real need explaining to the user why a given language is currently displayed. This is more of a debug information for the administrator, who is supposed to know what he's doing when using this option ;)

By removing this long message we also simplify the code. The only complication comes from the hook `i18n_content_translation_not_available` that can be used for a full customization. Here i keep an empty `$msg` for retro-compatibity (it used the contain the 2d part) but we could take the opportunity to review this filter.

### DIFF
--- a/admin/qtx_configuration.php
+++ b/admin/qtx_configuration.php
@@ -364,27 +364,27 @@ function qtranxf_conf() {
                                 <br/>
                                 <label for="hide_untranslated"><input type="checkbox" name="hide_untranslated"
                                                                       id="hide_untranslated"
-                                                                      value="1"<?php checked( $q_config['hide_untranslated'] ) ?>/> <?php _e( 'Hide Content which is not available for the selected language.', 'qtranslate' ) ?>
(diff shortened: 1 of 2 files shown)


herrvigg:
OK i managed to keep the original filter. I think all the filters should be reviewed and made consistent but it's another task. Anyway we should at least rename then together with the new plugin name.
I'm quite happy with this change now, i think it can be merged so more people can try this.

PR #659, opened 2019-03-26
Adds support for Composer

TimidRobot:
Adds support for [Composer](https://getcomposer.org/)


herrvigg:
OK i recently closed another PR for Composer because i found it confusing, pointing to another repo and mentioning qTranslate-X. This one looks better, it shouldn't cost much to have so let's take it. Anyway when we move to a new repo the project name will change so this one shouldn't conflict.

PR #673, opened 2019-04-23
Making embedded content translatable

Juje:
I noticed that embedded content wasn't translated so I looked how it could be done and my file change is the way how I succeeded into doing just that. I'm not sure where it should be placed so I added it in the `qtranslate_hooks` file.

Example link:
https://example.com/wp-json/oembed/1.0/embed?url=https%3A%2F%2Fexample.com%2Fpage%2F

### DIFF
--- a/qtranslate_hooks.php
+++ b/qtranslate_hooks.php
@@ -200,8 +200,4 @@ add_filter( '_wp_post_revision_field_post_title', 'qtranxf_showAllSeparated', 0 
 add_filter( '_wp_post_revision_field_post_content', 'qtranxf_showAllSeparated', 0 );
 add_filter( '_wp_post_revision_field_post_excerpt', 'qtranxf_showAllSeparated', 0 );
 
-add_filter( 'oembed_response_data', function ($data) {
-    global $q_config;
-    $data['title'] = qtranxf_split( $data['title'] )[$q_config['language']];
-    return $data;
-} );
+add_filter( 'oembed_response_data', 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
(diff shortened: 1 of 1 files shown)


herrvigg:
I don't know much about `oembed_response_data`. I think it's fine to add a hook there, but i'm just wondering is it only about the title to be translated?

Also i'm not sure `qtranxf_split` is the good function to call either, this is more for internal use. Can you try with `qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage` as most of the other filters do?

Juje:
I changed the callback to `qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage` and it worked like a charm. I also double checked the content but that was already translated.

herrvigg:
That looks good!
It can certainly be merged. Just two questions:
- since oembed_response_data is an array, the funny thing with `qtranxf_use` (which is called by the `qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage`) is that it will translate all keys. Do you know if the content was translated before this call or after?
- most (but not all) of the other filters use priority 0. Do you think it matters here?

Juje:
> That looks good!
> It can certainly be merged. Just two questions:
> 
> * since oembed_response_data is an array, the funny thing with `qtranxf_use` (which is called by `qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage`) is that it translates all keys of the array. Do you know if the content was translated before this call or after?
> * most (but not all) of the other filters use priority 0. Do you think it matters here?

Hmm I'm not sure what was translated before this to be honest (I could check it if you want me to :wink:).
And I'm also not sure if the prio would matter or not because for me it worked without it.

herrvigg:
Merged in master.

PR #678, opened 2019-04-29
ACF post object titles not translated

hencca:
… title

### DIFF
--- a/admin/qtx_admin_utils.php
+++ b/admin/qtx_admin_utils.php
@@ -281,23 +281,24 @@ function qtranxf_before_admin_bar_render() {
 }
 
 function qtranxf_admin_the_title( $title ) {
-	//todo this filter should not be used in admin area at all?
-	if ( defined( 'DOING_AJAX' ) && DOING_AJAX && !is_admin())//nav-menus.php#752
-	{
+	// For nav menus, keep the raw value as the languages are handled client-side (LSB)
+	if ( defined( 'DOING_AJAX' ) && DOING_AJAX && isset($_REQUEST['action']) && $_REQUEST['action'] == 'add-menu-item' ) {
+		// When a nav menu is being added it is first handled by AJAX, see "wp_ajax_add_menu_item" in ajax-actions.php
+		// For the call to the filter "the_title", see "wp_setup_nav_menu_item" in nav-menus.php
 		return $title;
 	}
 	global $pagenow;
 	switch ( $pagenow ) {
-		//case 'term.php':
-		//case 'edit-tags.php':
 		case 'nav-menus.php':
+			// When the nav menu is updated, keep the raw value.
 			return $title;
 		default:
+			// For general display purposes only
 			return qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage( $title );
 	}
 }
 
-add_filter( 'the_title', 'qtranxf_admin_the_title', 0 );//WP: fires for display purposes only
+add_filter( 'the_title', 'qtranxf_admin_the_title', 0 );
 
 //filter added in qtranslate_hooks.php
 if ( ! function_exists( 'qtranxf_trim_words' ) ) {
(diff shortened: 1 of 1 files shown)


admin/qtx_admin_utils.php
```diff
@@ -282,7 +282,7 @@ function qtranxf_before_admin_bar_render() {
 
 function qtranxf_admin_the_title( $title ) {
 	//todo this filter should not be used in admin area at all?
```
herrvigg: Wondering why this function even exists.

admin/qtx_admin_utils.php
```diff
@@ -282,7 +282,7 @@ function qtranxf_before_admin_bar_render() {
...
-	if ( defined( 'DOING_AJAX' ) && DOING_AJAX )//nav-menus.php#752
+	if ( defined( 'DOING_AJAX' ) && DOING_AJAX && !is_admin())//nav-menus.php#752
```
herrvigg: But this check is even more restrictive than the admin area, it's only for Ajax calls. Is it really what you intended?

herrvigg:
Shouldn't we simply remove the whole `qtranxf_admin_the_title` hook?

hencca:
> Shouldn't we simply remove the whole `qtranxf_admin_the_title` hook?

That would work, but I'm new to the code base, so I don't know why it was put there.

herrvigg:
I won't have much time this week to look at this, but next week i'll try to.

herrvigg:
@hencca sorry for such a long time but finally i tested this. I still have to better understand this old comment: 
https://github.com/qtranslate/qtranslate-xt/blob/a8aa6efe054bac7c1ce7a1391c76efbd839ce93d/admin/qtx_admin_utils.php#L283-L288

I think the comment only applies to the AJAX case and not the whole filter function. Adding the `!is_admin()` test as you did is essentially like removing the whole if statement. When we enter this function we should always have `is_admin() == true` (but not necessarily AJAX of course). Then it continues and returns the translated stuff.

The risk is that if some async processing requires the raw data then we might do something wrong here. In some cases the raw value might be necessary, especially if it is used by a function writing data in DB! So i tend to believe we rather need to handle the AJAX request for ACF more specifically, at least in a preventive way.

Subsidiary question (but not directly related): why is there this special case for nav-menus? Not sure why this is needed. Finally, when is `the_title` used in the admin pages (apart from this case)?

https://github.com/qtranslate/qtranslate-xt/blob/a8aa6efe054bac7c1ce7a1391c76efbd839ce93d/admin/qtx_admin_utils.php#L290-L297

herrvigg:
@hencca all right, i finally managed to setup xdebug with Docker to better understand what's going on! This change is more complicated than expected:
- in general we want to translate the title, the filter is used for example to display the list of posts.
- the nav-menus wants the raw values as they are handled on the client-side in Javascript (LSB).
- the reason for the AJAX check is [this fix in version 3.4.6.6](https://github.com/qtranslate/qtranslate-xt/commit/ba43331a00cf43e56f0fd5f7710d91c24abe68a8#diff-0456a6e2ce32ed87881f3e2a482e35fd) to handle the special case when a nav menu is added!

In that last case, an AJAX called is made and here we don't have the global page. So another filtering is needed, done here with DOING_AJAX. So we should keep this case. But i'd say for other AJAX calls i don't think we want the raw value. If we had the autosave feature that could be a problem, but this is disabled by qTranslate...

I think i found a good solution by adding a new check on the action itself with ` $_REQUEST['action'] == 'add-menu-item'`. I've sent a patch on your repo! This PR is updated directly so we should now be able to merge this. Note this was on the master branch (in general it's better to create separate fix branches).

herrvigg:
@hencca i made a small mistake when committing the file, adding the execution bit (due to a mess with WSL). I managed to remove it from the main repo but not on yours. Sorry about that. Ideally you should take the new master from the reference repo. If you have problems with this the best is just to delete the file and check it out again.

PR #680, opened 2019-05-05
update Hungarian language files to version 3.5.5

nb000:


herrvigg:
I was thinking maybe better go up to 3.6.0 but there are already other changes in the pipeline so let's merge this. Thank you for the update! Good to see new users / old users coming back :)

PR #690, opened 2019-05-14
Fixed reference errors for PHP 7.3 in frontend

sviluppomania:


herrvigg:
@sviluppomania i have to look at this more in deep. It can be problematic to remove references in some cases, there might be some side effects if the function was changing the underlying data.

herrvigg:
@sviluppomania before going further, I'm not sure about the motivation of these changes. What was the problem? I also use PHP7.3 and as far as i know there is no warnings left.

sviluppomania:
PHP 7.3. no longer uses & $ but only $

herrvigg:
> PHP 7.3. no longer uses & $ but only $

@sviluppomania To my knowledge there's no such change in PHP7.3 and not even in PHP7.x. You might be confusing with other warnings in PHP7.1 but that was because of inconsistencies with the WP callbacks, it's not the case here so it's another topic. I use PHP7.3 and have no issues though i'm not entirely sure i run all of this code.
Do you have an error message somewhere? In that case please send me the exact content.

sviluppomania:
I use the PHP Compatibility Checker plugin

herrvigg:
I know this one: https://wordpress.org/plugins/php-compatibility-checker/
But it only goes up to PHP 7.2 and i don't have such errors.

Can you give the exact content of the errors?

herrvigg:
@sviluppomania i close this as i don't have any updates.

I tried with https://wordpress.org/plugins/php-compatibility-checker/. It does not give such warnings and it does not handle PHP 7.3. so i don't understand what this is about.

This can be re-opened if new information is given.

PR #692, opened 2019-05-19
update Hungarian language files to version 3.6.0

nb000:
Also make some strings translatable in modules table header.


herrvigg:
Looks fine, thanks! I didn't know about the `_ex` function, good. We may generalize this.

PR #702, opened 2019-06-11
Update vendor info in i18n-config.json

sviluppomania:


i18n-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "vendor": {
-    "plugins/qtranslate-x": "3.4"
+    "plugins/qtranslate-xt": "3.4"
```
herrvigg: I don't think the vendor field is used somewhere so it's only informative. Not a big deal but can you change the version to "3.6"?

PR #740, opened 2019-10-05
Set a priority for the ACF/qTranslate format filter

henriqueccruz:
So it avoids a problem when using translatable fields on custom admin screen (ACF options page)
More on that here: https://www.advancedcustomfields.com/resources/options-page/
If not, priority 10 is going to be automatically assigned and fields will return empty.


herrvigg:
hello, thanks for this!
Just wondering why the priority 5 and not higher. Usually it's better to put the highest possible value.

Also, i understand you need something below 10 but from the link you gave i can't find any information about that. Which part were you pointing out?

henriqueccruz:
Because I ran into a very specific problem. The image field, when translated inside an option page (the page I linked above contains guidelines on how to create one) returned empty.
Setting the priority to anything less than 10 (9 for an example) solved the problem.

You can replicate the problem creating an option page (it works just like a theme option page, in the lateral menu), create a custom field item and a translatable image field inside.

Everything works as it should, until you try to fetch data to use, with functions like get_sub_field or get_field. It simply returns empty, which doesn't happen at all when using a normal image field.

Didn't confirm if it happens with other translatable fields in these conditions, though.

herrvigg:
OK i'll give it a try later just for double check. Anyway this will be merged soon.

herrvigg:
@henriqueccruz i couldn't try this because it requires ACF PRO which i don't have. But this looks fine, i just added an extra comment. Merged in master.

Issue #752, opened 2019-10-29
[WC] Product variations not show in product page


cyberneco:
Hi!

Congratulations for the project.

I'm creating a online store with Q Translate XT and Impreza Theme.

When i create a simple product all run correctly, but when i create a variable product, in the page of this, the dropdown list don't show the variations.

If i desactivate the Qtranslate plugin, it shows correctly.

An example: [](https://dev.sitenrain.ch/en/product/palmeri-navalta/)

I attach the report of the plugin.

Anybody can help me? 

Many thanks

[Report.txt](https://github.com/qtranslate/qtranslate-xt/files/3784133/Report.txt)

cyberneco:
For discarding some things: no is for CSS 

![photo_2019-10-29_22-06-22](https://user-images.githubusercontent.com/57144101/67809104-6223d800-fa98-11e9-8525-eab38ecfba4d.jpg)

herrvigg:
See also #612 #645

cyberneco:
Solved. I have installed the 3.5.5 version of Qtranslate-XT and 1.3 version of Woocommerce-Qtranslate.

Many thanks.

herrvigg:
It doesn't work with qT-XT 3.7.1? This is not solved in that case. These old versions are not going to be maintained and the more it goes the more you will have many other issues.

herrvigg:
Yes this error message is not very friendly, but the reason is that plugin is obsolete since qT-XT 3.6.0, there is now a specific module for that so you can't enable the old plugin later (to avoid the name conflicts more refactoring is needed). The module was created from the latest of this plugin and there's almost no specific change for WC since then (see release notes), so you should have find yourself in the same situation as before, plus all the fixes for qT-XT.

cyberneco:
> 
> 
> Yes this error message is not very friendly, but the reason is that plugin is obsolete since qT-XT 3.6.0, there is now a specific module for that so you can't enable the old plugin later (to avoid the name conflicts more refactoring is needed). The module was created from the latest version of this plugin and there's almost no specific change for WC since then (see release notes). So with the latest version (3.7.1), but without the obsolete plugin, you should find yourself in the same situation as before, plus all the fixes for qT-XT.

Ok. But, a solution for my problem is possible? Because i don't understand well the cause.

I have read the post #612 and #645 but i don't conclude anything

herrvigg:
[...]

Issue #612, opened 2018-09-17
[WC] Translating variable product attributes


wywarren:
Hi guys,

I made a similar post before on qtranslatex with no avail as well as on woocommerce's github so I will try to have a stab here.

This is my original post/hack solution on woocommerce:
https://github.com/woocommerce/woocommerce/issues/16623

As described, when we key product level multilingual attributes and apply them to variable products in woocommerce, the select box options disappears on the product page. If you go to view the source, they show up but I think at runtime, something in javascript or something else causes the variable production options to disappear.

I've tried making this work around in woocommerce source code base to get them to show up, but as described in the thread above, the returned object is an array although somehow it works in my implementation. I tried making a plugin to get around this using filters but have not come up with any ideas. I suspect this could be fixed by altering qtranslatex but I haven't had the chance to take a deep dive into the makings of qtranslatex to make a workable fix.

Any ideas or suggestions would help.

Thanks!

herrvigg:
hello,
thanks for reporting this.
My priority is currently Gutenberg but i keep this to be checked later when we can deal with WooCommerce. It's definitely one important plugin that qTranslate should support as best as possible.
I'll simply keep this open at the moment but we'll come back to it.

wywarren:
One work around that took me almost two days to pull off that I'm working with right now is I ended up making a plugin that subclassed the variable product class native to WooCommerce. From there, I copied over some functions to modify while others I just used the same functions as the parent variable product class. One of which was the function listed above where I replaced the function, effectively overriding it, in the new product type. I can post the plugin code if it helps. But it's quite a big fix while the original variable product is still broken for multiling strings. At least it works for our application though. I have not verified yet if it works with REST API as well (which is also what we're using it for).

herrvigg:
Let's see if we can integrate it directly in qTranslate-XT.

Actually some of the issues mentioned by other people should already be solved in the 1.4 version available on github, but they were never released to the official repo on wordpress.org still in 1.3. Same happened for qTranslate-X, strange if they were not directly synched for every release.

Anyway It makes no sense having a separate plugin for such a small thing so my plan is to integrate it as an add-on. Imo this should be the case for all the extra components to integrate major plugins such as WooCommerce. It makes the maintenance and evolutions so much easier and clearly these should be part of qTranslate-X-whatever-the-name.

[...]

PR #790, opened 2020-02-26
Fix WooCommerce variations error

PavloKrostylov:
Fixed showing WooCommerce attributes with error:
`Warning: implode(): Invalid arguments passed /wp-content/plugins/woocommerce/includes/wc-attribute-functions.php on line 41`

### DIFF
--- a/modules/woo-commerce/qwc-front.php
+++ b/modules/woo-commerce/qwc-front.php
@@ -105,12 +105,17 @@ function save_post_qwc_store_language( $order_id ) {
 }
 
 function qwc_product_get_attributes( $attributes ) {
-    //only 'value' needs to be translated at front end
     foreach ( $attributes as $key => $attribute ) {
-        if ( ! isset( $attribute['value'] ) ) {
-            continue;
-        }
-        $attributes[ $key ]['value'] = qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage( $attribute['value'] );
+		if(!empty($attribute['name'])) {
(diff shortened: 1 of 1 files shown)


herrvigg:
@PavloKrostylov many thanks for sending PR - much appreciated.

I'd like trying to reproduce the original problem. Can you describe the steps to create such product?

herrvigg:
@PavloKrostylov ok i managed to reproduce this! The fix solves some of the problems but there are others with the translations. Also, the method used here is through the `ArrayAccess` and the related offset functions. It seems the WC developers have changed this in favor of new methods, see [WC_Product_Attribute](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-product-attribute.php.
) especially the class comment:

> Uses ArrayAccess to be BW compatible with previous ways of reading attributes.

Therefore we should use the named get-set directly. I'm going to rewrite this a bit differently but now it's easy since you pointed out the problematic part.

herrvigg:
@PavloKrostylov did you try this with translations in the values of the custom attributes? I tried it and this removes all the variations having translation, because of a comparison done internally in WooCommerce.

Now i think i understand quite clearly the main two problems. This fix should only solve the `implode` issue #707 but not for the variations not shown #752. But since the solution for this second problem may require other changes i will not take the PR yet.

PavloKrostylov:
@herrvigg No. I fixed only the implode error, because variations did not work at all
So, please make some changes in qwc_product_get_attributes or totally disable this function because nobody can't use woocommerce variations with current version of qtranslate

herrvigg:
OK great news! I finally found a new solution with `woocommerce_dropdown_variation_attribute_options_args` so i close this PR. It works for custom attributes, with only some limitations for the edition. See #612 for more info.

Issue #198, opened 2018-06-22
Bug in qtranxf_wp_head()


herrvigg:
<a href="https://github.com/nb000"><img src="https://avatars2.githubusercontent.com/u/10193992?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [nb000](https://github.com/nb000)**
_Wednesday Jul 15, 2015 at 14:33 GMT_
_Originally opened as https://github.com/qTranslate-Team/qtranslate-x/issues/198_

----

Hi there!
I am writing about version **qTranslate-X 3.4.1**.
In the function `qtranxf_wp_head()`  where the meta tags are generated, you add the alternate language link tags. To this you call the `qtranxf_convertURL(...)` function.
As the last parameter is true, there could be URLs generated, which do not exists.

For example:
domain = example.com
active languages: hu, de, en
default language = hu
hide default language code from URL is set to true
On the URL http://example.com/de/somepageslug the following links are generated:

> &lt;link hreflang="hu" href="http://example.com/hu/somepageslug" rel="alternate" />
> &lt;link hreflang="de" href="http://example.com/de/somepageslug" rel="alternate" />
> &lt;link hreflang="en" href="http://example.com/en/somepageslug" rel="alternate" />
> &lt;link hreflang="x-default" href="http://example.com/sompageslug" rel="alternate" />

Here the first URL is invalid.
I think, the function should be called without the last parameter.

herrvigg:
<a href="https://github.com/johnclause"><img src="https://avatars2.githubusercontent.com/u/5491361?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [johnclause](https://github.com/johnclause)**
_Wednesday Jul 15, 2015 at 14:42 GMT_

----

It is by design and is ok with google. This url must switch the language to hu even for google robots, which do not use cookies. This document may help:  https://qtranslatexteam.wordpress.com/2015/02/26/browser-redirection-based-on-language/.

I am closing this issue for now, we can still write to closed issue and can re-open it if needed.

herrvigg:
<a href="https://github.com/nb000"><img src="https://avatars2.githubusercontent.com/u/10193992?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [nb000](https://github.com/nb000)**
_Thursday Jul 16, 2015 at 05:55 GMT_

----

Answer is correct, and accepted.
Nice writing, but the linked page misses a piece of information:
[...]

PR #819, opened 2020-04-15
hreflang tag to default language

antonzol:
If flag Hide URL language information for default language is active - hreflang to default language anyway include lang-pach.
Changing qtranxf_convertURL to defalult false - fix this problem

### DIFF
--- a/qtranslate_frontend.php
+++ b/qtranslate_frontend.php
@@ -48,7 +48,7 @@ function qtranxf_wp_head() {
             $hreflang = $lang;
         }
         //if($language != qtranxf_getLanguage())//standard requires them all
-        echo '<link hreflang="' . $hreflang . '" href="' . qtranxf_convertURL( '', $lang, false, true ) . '" rel="alternate" />' . PHP_EOL;
+        echo '<link hreflang="' . $hreflang . '" href="' . qtranxf_convertURL( '', $lang, false, false ) . '" rel="alternate" />' . PHP_EOL;
     }
     //https://support.google.com/webmasters/answer/189077
     echo '<link hreflang="x-default" href="' . qtranxf_convertURL( '', $q_config['default_language'] ) . '" rel="alternate" />' . PHP_EOL;


herrvigg:
I don't think we should this, see this discussion in https://github.com/qtranslate/qtranslate-xt/issues/198.
More explanations about the redirections and hreflang are given in a page that i moved to this wiki: https://github.com/qtranslate/qtranslate-xt/wiki/Browser-redirection

It is a bit difficult to understand but the reason to have the explicit language even with the option "hide default language" is to resolve the ambiguity from requests generated from SEO robots, despite the fact it generates a temporary redirection (i.e. the URL stored by the SEO is not a canonical one in that case).

herrvigg:
See #826: added a new filter `qtranslate_hreflang`. This leaves you the freedom to do whatever you want there. I propose to close like this.

herrvigg:
This topic is quite a complex question. I'm not sure about the right answer. Still need to think more about it...

herrvigg:
@antonzol I've reflected on this and i think the main behavior of qTranslate is correct in the default case. This is explained in the Wiki about [redirections](https://github.com/qtranslate/qtranslate-xt/wiki/Browser-redirection) but maybe it is easier to understand with an example. I will take the example from #198.

```
<link hreflang="hu" href="http://example.com/hu/somepageslug" rel="alternate" />
<link hreflang="de" href="http://example.com/de/somepageslug" rel="alternate" />
<link hreflang="en" href="http://example.com/en/somepageslug" rel="alternate" />
<link hreflang="x-default" href="http://example.com/sompageslug" rel="alternate" />
```
In this example the default language on the server is `hu` and we use the option "hide default language" (otherwise we would not have this problem).

The question is about the first link. You would want it to use the default URL without language (like the one of `x-default`) instead of the URL with `hu`.

Problem: qTranslate will normally negotiate the language the first time with your browser and store it in a cookie. The problem is the following:
- a "standard user" in hungary will get `hu`
- a "standard user" in germany will get `de`
- a german user living in hungary who prefers german will get `de` (browser settings) 
- and so on... with all possible combinations

The result is said to be non-deterministic: you don't know in advance what the result would be, because it depends on user parameters. Different results for different users.

So it would be wrong if the SEO assumes the default link is always `hu`. This would only be valid for the first user here but not the two others. So it is correct to use the `x-default` that should link to the default URL, despite the fact that maybe for you, you consider it to be only `hu`.

However - there is one special case mentioned by @nb000 in #198. If you disable the language negotiation (Languages / General / Detect Language Browser) you could assume the language is going to be fixed, therefore deterministic. In that case, the `hu` could be enforced to the default link because it would always be for that language.

Also one very extreme case would be if the user changes the cookie by external means, or for some reasons already have this cookie set (because the website had the auto detection first and removed it later). We may even consider the two conditions together (disable auto detect AND disable cookie) but maybe this is a bit of overkill.

Does this make it clearer?

herrvigg:
@antonzol i have now pushed a new change to fix the problem raised in #198 (very old ticket).

Description of the fix:
Only if the language is detected by the browser, enforce the language shown to make the default URL deterministic for SEO.
Otherwise, allow the option "Hide URL language information for default language".
Note the "x-default" item is not concerned by these changes, this link always uses the default URL with no override.

I have updated the Wiki with all these clarifications about the [redirections](https://github.com/qtranslate/qtranslate-xt/wiki/Browser-redirection) especially with a new clear section for the [SEO and hreflang](https://github.com/qtranslate/qtranslate-xt/wiki/Browser-redirection#localized-versions-for-seo-hreflang). 

So in short: if you want the behavior you asked in first hand, simply **disable the browser detection**. This keeps all the hreflang consistent and it requires no change anymore. Possibly don't use the new filter `qtranslate_hreflang` because i'm not sure i will keep it. I propose to close the PR with the fix (concerning #198).

herrvigg:
@antonzol can we close this?

herrvigg:
@antonzol i prefer to close this for now with the information given in this thread. You can still comment on this issue if you feel something is not clear.

herrvigg:
@antonzol fix released in 3.9.0 when the browser detection is disabled.

PR #827, opened 2020-04-26
FIX: Translate product name in WC order admin page

pjrobertson:
When viewing WooCommerce orders from the admin side, the name of the product ('item') wouldn't be translated properly.

Here's the issue in the admin (edit order) page:

Screenshot 2020-04-26 at 16 08 57

(P.S. I agree as herrvigg said that the Woocommerce file doing all this in PHP is messy, and should be changed to JSON. That's an issue for another time...)


herrvigg:
OK thank you. I think this worked previously but the recent fixes such as #612 could introduce small regressions here. Good it can be handled with just a display hook.

PR #831, opened 2020-04-30
Add CSS for language col in WC products page

### LINKED ISSUE #804: Woocommerce - Backend product list "language" column css to be fixed
As a minor bug, language column in admin product list has 0 width. Maybe a small css rule override could be implemented to reduce the width for one of the woocommerce default columns to let the language column in the table (es th.column-language {width: 5% !important;} )

pjrobertson:
Fixes #804

I'm not sure if this is the best way - adding custom CSS files to each module - but it seems like the most extensible going forwards. I see that the acf module also does this.

### DIFF
--- a/modules/woo-commerce/qwc-admin.css
+++ b/modules/woo-commerce/qwc-admin.css
@@ -1,0 +1,3 @@
+.fixed .column-language {
+        width:10%;
+}
--- a/modules/woo-commerce/qwc-admin.php
+++ b/modules/woo-commerce/qwc-admin.php
@@ -45,6 +45,12 @@ function qwc_add_filters_admin() {
 
 qwc_add_filters_admin();
 
+add_action( 'admin_enqueue_scripts', 'qwc_add_admin_styles' );
+function qwc_add_admin_styles() {
+        wp_register_style( 'qwc_qtranslate_admin', plugins_url( '/assets/qwc-admin.css', __FILE__ ), array(), QTX_VERSION);
+        wp_enqueue_style( 'qwc_qtranslate_admin', plugins_url( '/assets/qwc-admin.css', __FILE__ ), array(), QTX_VERSION);
+}
+
 add_filter( 'qtranslate_load_admin_page_config', 'qwc_add_admin_page_config' );
 function qwc_add_admin_page_config( $page_configs ) {
     // post.php


modules/woo-commerce/qwc-admin.css
```diff
@@ -0,0 +1,3 @@
+.fixed .column-language {
```
herrvigg, line 1: Where does CSS `column-language` come from?

herrvigg:
OK got it, it's about the products not the orders :)

herrvigg:
This fixed layout CSS in the `wp-list-table` is really bad. As soon as the table is reduced the column is messed up again. Not flexible or responsive at all 😒 But this doesn't come from qtranslate.
We need a better solution than the relative width (%).  It's already the case for cols name, SKU, product category, product tag. I think they should remove this in WC. But we should not add a new % here.

herrvigg:
I'm not a web designer but this doesn't look good. Imo an issue should be opened on WooCommerce to remove these % because they don't live well with the `table-layout: fixed`. Such col sizes with % would be computed relatively to the current size of the table but then if you resize the page it's a mess. Said differently it's not even that it's not responsive design, it's anti-responsive, making it worse.

I think even that the fixed layout itself is a bad practice, it's only meant to make the computations faster maybe for mobile but in 2020 i don't think i makes much sense. This comes from WP so it's certainly even harder to change. But can be done independently i believe. The first thing is certainly to ask WC to switch to really-fixed size such as `px`.

For qTranslate we should do something similar, avoid % and used a fixed size instead.

pjrobertson:
> This fixed layout CSS in the `wp-list-table` is really bad. As soon as the table is reduced the column is messed up again. Not flexible or responsive at all 😒 But this doesn't come from qtranslate.
> We need a better solution than the relative width (%). It's already the case for cols name, SKU, product category, product tag. I think they should remove this in WC. But we should not add a new % here.

I'm not having this issue. Have you checked resizing the table on a live install? Even with this change, the table is responsive and works for me. Below 783px, the table goes into 'compact' mode. The `width:10%` CSS was using the convention that's already in use in WooCommerce (see the `column-date` element). I agree, it's not ideal, but there's not much point us changing the CSS to something better here if the rest of the WC table doesn't follow the same trend.

This is a fix to an already existing, big problem. My goal here is to just get things working so they are actually usable.

Before:

Screenshot 2020-05-06 at 16 18 29

After:

Screenshot 2020-05-06 at 16 18 48

modules/woo-commerce/qwc-admin.css
pjrobertson: Sorry, it's the WooCommerce products page, apologies for the incorrect title

herrvigg:
All right, i'm not super happy with the non-responsive layout but it is indeed depending on WC and they would have to improve it for the whole table. At least we are consistent with them and if they change it we can adapt later.
Thanks for the PR!

herrvigg:
There was a wrong path for the new CSS stylesheet. Fixed in #894.

Issue #853, opened 2020-06-17
PHP Warning Cannot modify header information - headers already sent in qtranslate_core.php on line 463


DevWael:
Hello there...
The plugin raises a PHP warning 
```php
PHP Warning:  Cannot modify header information - headers already sent in /.../wp-content/plugins/qtranslate-xt-3.8.1/qtranslate_core.php on line 463
```

herrvigg:
hello,
there's nothing i can do with so little information. You have to debug it yourself and provide more relevant data about the issue you have. Check the [Troubleshooting section](https://github.com/qtranslate/qtranslate-xt/wiki/Troubleshooting).

DevWael:
thank you, I will debug it and get back to you with more info

DevWael:
hello,

here are the details of my installation website:
WordPress: 5.4.2
PHP: 7.1.33
Database: Mysqli 5.7.30
qTranslate-XT: 3.8.1 

I have also seen this error in the same function in previous Qtranslate-X versions.

the error related with this function:
```php
function qtranxf_setcookie_language( $lang, $cookie_name, $cookie_path ) {
    global $q_config;

    // only meant for server-side, set 'httponly' flag
    setcookie( $cookie_name, $lang, strtotime( '+1year' ), $cookie_path, null, $q_config['use_secure_cookie'], true );
}
```

if you still need more information you are welcomed

herrvigg:
I close this issue, the warning is actually a "good thing" to signal a bug. You may hide it on production site but this hides a real problem. There is something else writing on the main output before the headers, making `setcookie` ineffective, which is bad.
Maybe what you can do is raise an exception here and/or check what you get before the warning. The goal is to find which component is writing, it's quite unlikely that it is qTranslate.

PR #856, opened 2020-06-28
hide the error related with the issue #853

DevWael:


herrvigg:
You should not hide the error. The problem is some other output is sent before `setcookie` which makes it ineffective. The warning is actually a good thing you want to see because it forces you to fix the real underlying issue.

Issue #433, opened 2018-06-23
Stop forcing suppress_filters to false on WP_Query


herrvigg:
<a href="https://github.com/barryceelen"><img src="https://avatars3.githubusercontent.com/u/4179791?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [barryceelen](https://github.com/barryceelen)**
_Thursday Sep 22, 2016 at 23:00 GMT_
_Originally opened as https://github.com/qTranslate-Team/qtranslate-x/pull/433_

----

The `qtranxf_pre_get_posts` action forces `$query->query_vars['suppress_filters'] = false;`. This way   on the front end no queries can be performed using WP_Query where the `suppress_filters` argument is set to true (unless one removes the action altogether before performing a query with WP_Query). 

----

_**[barryceelen](https://github.com/barryceelen)** included the following code: https://github.com/qTranslate-Team/qtranslate-x/pull/433/commits_

herrvigg:
See https://github.com/qtranslate/qtranslate-xt/pull/872#issuecomment-695785393

We can't just remove this without risking other regressions. I don't know how to fix this, I close this until someone understands this better and proposes a better solution.

PR #872, opened 2020-08-21
Removes enforcement of suppress filters=false

Bones5:
https://github.com/qTranslate-Team/qtranslate-x/pull/433/commits/b8146a1be459bd7b08ec6ace3ff7bb2954a29511

https://github.com/qtranslate/qtranslate-xt/issues/433

### DIFF
--- a/qtranslate_frontend.php
+++ b/qtranslate_frontend.php
@@ -579,22 +579,6 @@ function qtranxf_postsFilter( $posts, $query ) {//WP_Query
 
 add_filter( 'the_posts', 'qtranxf_postsFilter', 5, 2 );
 
-/** allow all filters within WP_Query - many other add_filters may not be needed now? */
-function qtranxf_pre_get_posts( $query ) {//WP_Query
-	//qtranxf_dbg_log('qtranxf_pre_get_posts: $query: ',$query);
-	//'post_type'
-	if ( isset( $query->query_vars['post_type'] ) ) {
-		switch ( $query->query_vars['post_type'] ) {
-			case 'nav_menu_item':
-				return;
-			default:
-				break;
-		}
-	}
-}
-
-add_action( 'pre_get_posts', 'qtranxf_pre_get_posts', 99 );
-
 /**
  * since 3.1-b3 new query to pass empty content and content without closing tags (sliders, galleries and other special kind of posts that never get translated)
  */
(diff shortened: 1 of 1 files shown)


Bones5:
Ok yes I see that, PR rewritten.

Yes it probably was there for a reason, I will try and investigate. 

It appears to just re-enforce the default and prevent you from using `suppress_filters` in your code. This can make it difficult to select the language you want programatically, I currently want to construct an object with both languages to pass to a search index. I have had similar issues trying to pass an object to JS for animation

Bones5:
I have managed to do a bit of testing on this today, so far I have found one effect of the change:

When using `get_field` on an ACF Post Object the `post_title` is now returned without filters. This is actually a more consistent response as `post_content` has always been returned without filters being applied, we typically use `apply_filters` on the content but not on the title.

There doesn't appear to be any change in the behaviour of a standard `get_posts()` or `WP_Query`, I'm unable to understand why `get_field()` is not working in the same way so cannot offer a fix for backwards compatibility.

This does currently make this a breaking change but again provides a more consistent experience that allows the developer to apply filters at their discretion.

I still believe this is a worthwhile change as given the progress towards JS and React in WordPress, the ability to pass multiple translations of strings to services outside of PHP and the current language is becoming more useful. It also prevents the undesired override of a standard WordPress feature

herrvigg:
The specific case for the `nav_menu_item` came in qTranslate-X release 3.3.3. I found some discussions about this `pre_get_posts` filter here:
https://wordpress.org/support/topic/menu-conflict-with-plugin-custom-field-template/page/3/

> commenting out 331, cannot be a permanent solution, since it will break many other sites, although it is a very valuable information to know that it conceals your problem.

The row number changed but for sure it is equivalent to remove this `suppress_filter=false` override. I don't really understand what was meant with "it will break many other sites"... 😕

Bones5:
I've realised that the `post_content_ml` and `post_title_ml` fields actually perform the function I'm looking for.

It feels that this function could be ensuring compatibility with other plugins and there are too many unknown effects for it to be removed

herrvigg:
Great you find a way to solve this! 
I'm still not sure about the effect of removing the filter override entirely. Once we move to the next repo i will try to remove it and restart from scratch, we'll see better what we really need to keep.

PR #880, opened 2020-09-05
Extend language code format

### LINKED ISSUE #668: hreflang should support ISO 639-1, ISO 3166-1 Alpha 2, ISO 15924 formats
According to Google: https://support.google.com/webmasters/answer/189077

> The value of the hreflang attribute identifies the language (in ISO 639-1 format) and optionally a region (in ISO 3166-1 Alpha 2 format) of an alternate URL. (The language need not be related to the region.) For example:
>
> - de: German language content, independent of region
> - en-GB: English language content, for GB users
> - de-ES: German language content, for users in Spain

Since some languages have different type, such as Chinese Tranditional and Chinese Simplified, ISO 639-1 format only describes 1 language code "zh" for both Chinese languages, but we cannot fill "zh" for both Chinese Tranditional & Chinese Simplified at the same time.

From Google answer, we should use "zh-TW" / "zh-CN" or "zh-Hant" / "zh-Hans" instead, but the "Language Code" field in qtranslate-XT setting limit 2 characters only.

Could qtranslate-XT release the 2 characters limit so that we can follow what Google suggests?

herrvigg:
Fixes #668.
Language code format: `language[-script][-region]`
- language ISO 639-1 and 639-2: 2 or 3 alpha
-  script ISO 15924: 4 Alpha
- region ISO 3166-2: 2 ALPHA or 3 digits


herrvigg:
After comments in #668 limit extension to ISO 639-1 and 639-2: 2 or 3 alpha for now.
Let's first release these changes because it touches many core parts and the discussion goes on for the other parts.

qtranslate_core.php
```diff
@@ -336,7 +337,7 @@ function qtranxf_parse_language_info( &$url_info, $link = false ) {
         } else if ( isset( $_POST['lang'] ) ) {
             $query_lang = qtranxf_resolveLangCase( $_POST['lang'], $doredirect );
         }
-    } elseif ( ! empty( $url_info['query'] ) && preg_match( '/(^|&|&amp;|&#038;|\?)lang=([a-z]{2})/i', $url_info['query'], $match ) ) {
+    } elseif ( ! empty( $url_info['query'] ) && preg_match( '/(^|&|&amp;|&#038;|\?)lang=($lang_code)/i', $url_info['query'], $match ) ) {
```
herrvigg: Regression - the regex should be double quoted so that the new `$lang_code` can be evaluated.

Issue #794, opened 2020-03-05
Yoast module disabled


herrvigg:
For the history: the Yoast module was initially created from the the old integration plugin [Yoast SEO & qTranslate-X](https://github.com/qTranslate-Team/wp-seo-qtranslate-x).

Due to many problems with the existing JS code that had not been maintained for too long, also because the PHP part wasn't really relevant, it was preferable to disable the Yoast module for now, as [described here](https://github.com/qtranslate/qtranslate-xt/issues/661#issuecomment-479661609). 

Developers willing to re-activate those features are much welcome. If some PHP components are relevant we may reactivate the module for those and only keep the JS part disabled.

herrvigg:
I will try to re-enable the module because i realized it's actually doing useful things in the admin section. The problem is more the javascript part that needs more work and i don't think i will have much time to look into all the details, but let's do it progressively.

ashabelnikov:
Great news!

mohjak:
@herrvigg I would like to help in this regard, because I am facing issues in integrating qtranslate-xt with YoastSEO, for example,  the metadata like `og:description` appears with the language code [:en].

```html
<meta property="og:description" content="[:en]Around 1.5 million people will be displaced in the Sundarbans, and the process has started[:bn]সুন্দরবন সংলগ্ন এলাকাসমূহে বসবাসকারী প্রায় ১.৫. মিলিয়ন মানুষ বাস্তুচ্যুত হয়ে পড়বে অচিরেই, আর এরই মধ্যে শুরু হয়ে গেছে এই প্রক্রিয়া[:]" />
```

So that, where do you think it is better to start from. I have already started to review the source code of [https://github.com/qTranslate-Team/wp-seo-qtranslate-x](https://github.com/qTranslate-Team/wp-seo-qtranslate-x). 
I would like to know if you have any plan regarding this issue.

pjrobertson:
Yoast have just come out with a new update (14.0) which they claim has [overhauled the backend](https://yoast.com/yoast-seo-14-0/). They state:

> We’ve also found that not all plugins that rely on Yoast SEO have updated to our 14.0 code yet. If you use, for instance, a multilingual plugin, please check that they’re ready for the update.

So I suggest this module be fixed with Yoast 14.0 in mind as opposed to previous versions.

pjrobertson:
[...]

PR #892, opened 2020-09-25
Fix Javascript check for Yoast

pjrobertson:
Fix bug with certain plugins where this code loads but YoastSEO isn't available.

Some pages seemed to be loading the qwpseo-exec script, even though Yoast wasn't being loaded.

I'm not quite sure of the underlying issue, but this is a quick fix that will solve this problem short of the complete re-coding that the Yoast module needs. This just checks that window.YoastSEO exists first.


herrvigg:
> Some pages seemed to be loading the qwpseo-exec script, even though Yoast wasn't being loaded.

In fact that doesn't sound good, this script should not be loaded at all in that case. Will check better to see if i find something.

herrvigg:
@pjrobertson but currently the Yoast module is still disabled. How did you enable it again, just uncommented the related block in qtx_modules_handler.php? You should also deactivate/activate qTranslate-XT, otherwise the Yoast plugin status might not be updated.

pjrobertson:
Yes, that's what I did @herrvigg. Noted on the deactivate/activate process, it's been working find for the past 6 months!

herrvigg:
In fact, this is not normal. The script should never be called if the Yoast script is not loaded, because there is dependency to `yoast-seo-post-scraper` or `yoast-seo-term-scraper`. 

The fix should be fine to get around this but I would like to be able to reproduce this to understand. Could you try to isolate the case (minimum number of plugins) and the page where this happens?

herrvigg:
@pjrobertson any update?

herrvigg:
Let's merge this change, it doesn't cost much. Anyway more work is needed to make Yoast work again, see #794.

herrvigg:
Merged after fixing the conflict locally.

PR #897, opened 2020-09-29
Placeholder goes over the title when adding a new post

vincequeiroz:
When creating a new post and changing from one language to another the placeholder goes over the title.

The title's input field class, when I'm typing a text, is "screen-reader-text", but when I change from one language to another it changes to "title-prompt-text".

image

The problem is that the code uses the function .attr(...), but this function is only getting the objects value at the start (when the html is created). The correct function is val(), because it's necessary to get the object's property value, which can change many times.

image


bagaweb:
Thanks @vincequeiroz I noticed the same problem.

herrvigg:
Thank you!
It's funny because i also noticed that problem briefly but now i'm not able to reproduce it! Does this happen before you create the post? I made some quick test before the patch, `attr('value')` and `val()` give me the same value (updated) which is surprising.

herrvigg:
We can certainly take this small fix even preventively, because `val()` is the right way to do.
But still this puzzles me and might hide some other problem... i saw this problem before and i'm not able to reproduce it even once. I even tried with different browsers, same. Was that due to some combination with other plugin or a previous error that prevented the update to be done? If you can describe exactly the situation when this happen step by step, i could understand better what is behind.

vincequeiroz:
Hi @herrvigg 

When I debugged the code, I couldn't see other plugins interacting with the post's title.

For me, the error happen when I add a new Post, type a title and change from one language to another.

I'm using the Wordpress 5.5.1 with jQuery 1.12.4.

Screen Recording 2020-10-05 at 10 23 36

herrvigg:
OK thank you! I still don't manage to reproduce it... but let's take your fix. The code is better like this.

PR #910, opened 2020-10-29
Fix paragraph deleting in classic ACF WYSIWYG

antonzol:
Before in classic ACF WYSIWYG paragraphs was deleting.
This make classic ACF field work korrectly with qtranslate-xt installed.

### DIFF
--- a/admin/js/common.js
+++ b/admin/js/common.js
@@ -649,7 +649,7 @@ function qtranxj_ce(tagName, props, pNode, isFirst) {
 
         var updateTinyMCE = function (h) {
             text = h.contentField.value;
-            if (h.wpautop && window.switchEditors) {
+            if (1) {
                 text = window.switchEditors.wpautop(text);
             }
             h.mce.setContent(text, {format: 'html'});


admin/js/common.js
```diff
@@ -649,7 +649,7 @@ function qtranxj_ce(tagName, props, pNode, isFirst) {
 
         var updateTinyMCE = function (h) {
             text = h.contentField.value;
-            if (h.wpautop && window.switchEditors) {
+            if (1) {
```
herrvigg, line 652: You are removing the if-check without removing it.
About the 2 conditions:
- why do you remove the check h.wpautop? 
- why do you remove the check on window.switchEditors? It is still used in your code.

herrvigg (commented):
I don't understand this change.

> Before in classic ACF WYSIWYS paragraphs was deleting.
> This make classic ACF field work korrectly with qtranslate-xt installed.

Do you mean paragraphs are deleted?

admin/js/common.js
antonzol: In "classic" WYSIWYG editor with ACF and qtranslate-xt wordpress does not use wpautop().
And it makes paragraphs and new line deleted. during opening editor.

This make wpautop works anyway

herrvigg: I'm not sure about these changes, more investigation and testing is needed to understand this better. I would also like WP to fix the issue with Firefox #931.

herrvigg: Not that i don't want to merge the fix, but the change would need to be reworked a bit. If the if-check is not necessary it should be completely removed but i feel it was there for good reasons.

herrvigg: wpautop is disabled in the qTranslate PHP code but not in the JS code, see `updateTinyMCE` in `admin/common.js` and all the comments from 3.2.9.8 "ugly trick" about `updateTinyMCEonInit`.
I don't think we can disable it like this. Maybe we need a special case for ACF WYSIWYG but I would not do it in all cases.

antonzol:
> @antonzol I understand better now why we have this `wpautop` check in the JS code. The `wpautop` is actually supported ~for single language mode~ after translation of the ML content. There might be something broken for ACF though.
> 
> Can you explain me exactly the problem so I can reproduce it?
> 
> * Please describe all the steps. What content you enter, what operations you do, what you get and what you'd expect.
> * There are two WYSIWYG editors: Wysiwyg Editor and Wysiwyg Editor (qTranslate). Which one are you using?
> * Is it in Visual mode or Text mode?

I do not remember exactly.
I was in all projects with ACF and qtranslate.
But i think now it is fixed.

herrvigg:
@antonzol I propose to close this PR.
I pushed a change to remove the `wpautop` field from QT-XT hook fields, see #1024.
This just shows we apply the `wpautop` settings from TinyMCE but does not change anything for QT-XT.
In general, `wpautop` is set to `true`.

I made a few checks and in fact some paragraphs are removed if you enter `<p>some text</p>` in text mode and then switch to visual mode. However, this happens even if QT-XT is disabled, it's the normal behavior of WordPress for any TinyMCE editor. It's the case for the main content editor and ACF WYSIWYG, they work the same.

herrvigg:
From my understanding the main philosophy is this:
- in text mode, if you enter two lines breaks it becomes a paragraph like `<p>`. One like break is just like a `<br>`. But you should not really save these raw tags as such, WordPress removes them as soon as you switch to Visual Mode so it's better to stick to double or single newline.
- in visual mode, if you type a new line with the "Enter" key, it means new paragraph. If you want just a single linebreak type "Shift + Enter".

Issue #675, opened 2019-04-23
Different featured image per language


tuncbz:
I only have the possibility to add one featured image in blog posts. How to change the featured image in other languages?
Thank you !

herrvigg:
As far as i know such feature is not yet supported. It would be interesting to have for sure.

We should keep better track of this. There used to be an old list with desirable features coming from the [legacy website](https://qtranslatexteam.wordpress.com/desirable/).

I'm currently migrating this website to wiki pages but it requires quite some work to put some order and update it with the current state. This is also a preliminary step before doing the migration to a new plugin, i first want this repo entirely self-contained without any more external reference.

PR #937, opened 2021-01-12
new features

nabi009:
Would you please add translation language feature to featured image or (thumbnail). mean thumbnail should be translatable mean we have different image for different language.


herrvigg:
The code change you propose has nothing to do with what you are requesting about the image thumbnail. The code you are propose is my own change to handle regional codes and scripts in the language format (#668). This requires other changes so that's why i didn't merge it yet. 

The request for the feature image exists already, see #675. Unfortunately i don't have time to work on this and i'm almost the single developer on this project, which i only do on my free time as a hobby.

So I propose to close this PR.

nabi009:
thanks so much, dear @herrvigg, yes that is your code and I did not suggest any code just I request the new features. anyhow thanks.

PR #945, opened 2021-01-19
Redesign language list with icons

herrvigg:
New design for the language list:
- icons replace the text descriptions, but those are still available when hovering the icons
- add column for the locale (the one for WordPress, this might evolve)
- add column for the status (filled star = default, empty star = enabled, nothing = disabled)

image


herrvigg:
I merge this in master so everyone can try it directly and comment. This is easy to change.

Issue #755, opened 2019-11-08
Woocommerce products name in widgets


adryxxx:
Hello and congratulations for the excellent work!
I used qtranslatex and found an error in the woocommerce widgets: the title of the product can be seen with the language tags.
Example: [: it] Red [: en] Red [:]

I replaced the plugin with the latest version of qtranslateXT but the problem is always there.
I did various tests to see if it is the theme but with any theme the result does not change.
Has anyone solved the problem?

WP version 5.2.4
PHP version 7.2.22

Thanks in advance!
Adriano

herrvigg:
The field is probably missing in the selectors for finding the translatable fields. Usually the integration is done through a JSON file (which i dislike) but in the case of WooCommerce it's done with the built-in module, here for the admin section: https://github.com/qtranslate/qtranslate-xt/blob/master/modules/woo-commerce/qwc-admin.php

You should find how to select this field and add it here. If it's only for display and not edition, add encoding = display.

More info about the features can be found in the specification of the JSON format (the PHP code is doing the same): https://github.com/qtranslate/qtranslate-xt/wiki/JSON-Configuration

adryxxx:
Thanks for the reply!
I identified the id of the field and inserted the code in the file you told me but the result does not change. Did I do something wrong?

$fields['product-title'] = array( 'jquery' => '.product-title', 'encode' => 'display' );

But in the front file should I insert something?

![screenshot-www vitaben it-2019 11 12-11_17_39](https://user-images.githubusercontent.com/57527637/68662956-1cf2b200-053e-11ea-84f9-b003dc0c5964.jpg)

herrvigg:
But did you add this field on the right form? First you need to select the right form, then the fields. For the widgets i don't know exactly where it is though.

The front file is only for the "normal" rendering of the pages out of the admin section.

adryxxx:
Looking good I noticed that unfortunately there is no specific widget module. I wouldn't know where to start. :(

herrvigg:
Currently the modules are only meant to support the integration with a specific plugin, not for handling the widgets. For those you might actually take a look at the JS part: https://github.com/qtranslate/qtranslate-xt/blob/master/admin/js/widgets-exec.js

Sorry, i won't have much more time for this.

adryxxx:
OK thanks anyway.

adryxxx:
Allow me a consideration:
[...]

PR #957, opened 2021-02-05
[WC] Add front filter for product_get_name

erichk4:
Add the 'woocommerce_product_get_name' filter to the $use_filters array to translate the product name in the woocommerce widgets


erichk4:
Side note: it should solve #755, but in general only works for the non-cached woo widgets (e.g. Recent Products Widget)

PR #960, opened 2021-02-09
added Bangla localization files

zenideas:
I have added Bangladeshi Bangla translations. About 100 strings not done yet though. But, I'll try to complete those soon also.


herrvigg:
Great to have some updates on the localizations!

I was thinking about this recently and it feels the settings panels have way too much text, it's too verbose and not so meaningful for translations. Some parts are technical details that should rather go to Wiki pages.

I may start to simplify this soon so perhaps wait just a little bit with the longest strings in the settings.  The short ones will certainly not change so you could perhaps start with those instead.

zenideas:
> I was thinking about this recently and it feels the settings panels have way too much text, it's too verbose and not so meaningful for translations. Some parts are technical details that should rather go to Wiki pages.
Nice. So large texts will not be inside codes? I am not sure how the plan is. But would like to watch.
> 
> I may start to simplify this soon so perhaps wait just a little bit with the longest strings in the settings. The short ones will certainly not change so you could perhaps start with those instead.
As I mentioned earlier. When there are more than one '%s' in a sentence, there are possiblities the sequence of them would be hard to maintain in most languages. I pasted a link from php forums in the other conversation with you. The idea to add numbered attributes like '%1s', '%2s' and so on...

herrvigg:
The reference strings for the translations may evolve but let's take your current changes.

Issue #886, opened 2020-09-16
Flag Disable language client cookie "qtrans_front_language" does not work.


antonzol:
I use Pre-Path Mode (Default, puts /en/ in front of URL). SEO friendly.  URL
And 
Hide URL language information for default language.

And in some cases (30%) when I change language from language switching buttons to default - it automatically redirect me to last language, which is set in qtrans_front_language.

Checking or uncheking _Disable language client cookie "qtrans_front_language"_ doesn help.

When I clear qtrans_front_language cookie and after swith lenguage - everything is OK.
 But in day or two using website problem come back.

Maybe there something in LSB tracking wrong.

Another users have same on this website.
versions:
`{
  "PHP_VERSION": "7.3.20",
  "WP_VERSION": "5.5.1",
  "QTX_VERSION": "3.9.0",
  "Plugins": [
    "Додаткові поля Pro 5.9.1",
    "Classic Editor 1.6",
    "Cloudflare 3.5.1",
    "Contact Form 7 5.2.2",
    "Cyr to Lat enhanced 3.5",
    "Flamingo 2.2",
    "LiteSpeed Cache 3.4.2",
    "qTranslate-XT 3.9.0",
    "Simple Sitemap 3.5",
    "Yoast SEO 13.4.1",
    "WP Force Lowercase URLs 2.0.1",
    "WP Super Cache 1.7.1"
  ]
}`

configuration
`{
  "default_language": "uk",
  "enabled_languages": [
    "uk",
    "en",
    "ru"
  ],
  "flag_location": "plugins/qtranslate-xt-master/flags/",
  "language_name": {
    "uk": "ua",
    "en": "En",
    "ru": "Ru"
  },
  "locale": {
    "uk": "uk",
    "en": "en_US",
    "ru": "ru_RU"
  },
  "locale_html": {
    "uk": "",
    "en": "",
    "ru": ""
  },
  "not_available": {
    "uk": "Вибачте цей текст доступний тільки в “%LANG:”, “: і “%”.",
    "en": "Sorry, this entry is only available in %LANG:, : and %.",
    "ru": "Извините, этот текст доступен только на “%LANG:”, “:” и “%”."
  },
  "date_format": {
    "uk": "%A %B %e%q, %Y",
    "en": "%A %B %e%q, %Y",
    "ru": "%A %B %e%q, %Y"
  },
  "time_format": {
    "uk": "%H:%M",
    "en": "%I:%M %p",
[...]

PR #1008, opened 2021-04-16
Fixed issue with qTranslate not loading after clicking on the "Edit" button in the frontend

mweimerskirch:
We encountered an issue where qTranslate would not load when a user clicks on the "Edit" button in the frontend. This is caused by the check on the referer to see if the previous URL contains "wp-admin". This is obviously not the case when you come from the frontend directly to the edit page in the backend. I modified this check, so that it also checks if the currently called page contains "wp-admin".

### DIFF
--- a/qtranslate_core.php
+++ b/qtranslate_core.php
@@ -165,11 +165,12 @@ function qtranxf_detect_language( &$url_info ) {
     // parse language and front info from HTTP_REFERER
     if ( isset( $_SERVER['HTTP_REFERER'] ) && $parse_referrer ) {
         $http_referer             = $_SERVER['HTTP_REFERER'];
+        $http_request_uri         = $_SERVER['REQUEST_URI'];
         $url_info['http_referer'] = $http_referer;
 
         // if needed, detect front- vs back-end
         $parse_referrer_language = true;
-        if ( strpos( $http_referer, '/wp-admin' ) !== false ) {
+        if ( strpos( $http_request_uri, '/wp-admin' ) !== false || strpos( $http_referer, '/wp-admin' ) !== false ) {
             $url_info['referer_admin'] = true;
             if ( ! isset( $url_info['doing_front_end'] ) ) {
                 $url_info['doing_front_end'] = false;


qtranslate_core.php
```diff
@@ -165,11 +165,12 @@ function qtranxf_detect_language( &$url_info ) {
     // parse language and front info from HTTP_REFERER
     if ( isset( $_SERVER['HTTP_REFERER'] ) && $parse_referrer ) {
         $http_referer             = $_SERVER['HTTP_REFERER'];
+        $http_request_uri         = $_SERVER['REQUEST_URI'];
         $url_info['http_referer'] = $http_referer;
 
         // if needed, detect front- vs back-end
         $parse_referrer_language = true;
-        if ( strpos( $http_referer, '/wp-admin' ) !== false ) {
+        if ( strpos( $http_request_uri, '/wp-admin' ) !== false || strpos( $http_referer, '/wp-admin' ) !== false ) {
```
herrvigg, line 173: That part of the code is dedicated to parsing HTTP_REFERER ($parse_referrer = true). Not sure we should look at the request URI here.

herrvigg (commented):
I need to understand better that case. In first place, why are we parsing the referrer? 
That means the cookie is not set and the lang is not specified. Do you have the cookies enabled?

mweimerskirch:
Sorry, I just noticed today that I had to reply here when the issue re-appeared on another project.

Steps to reproduce:
1. Log into a WordPress site (editor or higher access level).
2. Browse to any regular page on the site.
3. Click on "Edit page" to open the current page in the editor.
4. The editor will look broken (the two languages are below one another), as qTranslate did not initialize correctly.

This is caused, because qTranslate-xt checks where the user comes from, using the referrer. If they come from the pages list in wp-admin, everything works fine. But if you come from the frontend, the editor breaks, because qTranslate thinks you're trying to load a frontend page. 

I fixed it (or worked around it) by not only checking the referrer, but also the currently loaded page.

Not sure why this check was there in the first place though.

qtranslate_core.php
herrvigg: @mweimerskirch this check should probably not be in the if-statement regarding HTTP_REFERER. How can we reproduce the problem you encountered?

herrvigg: My bad, I missed a few things... I'll check this again better.

herrvigg:
@mweimerskirch Sorry for late answer. I somehow understand the problem but I don't manage to reproduce it. In my case, I manage to go from front->admin to edit any page or post without problem.

When we enter this check on HTTP_REFERER, it means QTX did not manage to detect the information about the language yet. So it tries to find it in the HTTP_REFERER. In my case, it works because the cookie is set so we don't need to check the referrer at all. There is a known issue #886 if the cookies are disabled and this topic seems related. Can you first confirm you have the cookies enabled?

herrvigg:
@mweimerskirch I close this as there's no answer from your side. I don't manage to reproduce the problem and I don't really understand it, so we'd need complementary information from you, but it could probably be fixed by PR #1324.

Issue #1000, opened 2021-04-07
Rework JS integration without code injection


herrvigg:
A major rehaul of the JS part is currently going on with webpack and babel. I'm realizing now there are a few issues with the `i18n-config.json`:
- it allows to embed JS code directly. Code injection like that is not safe, we should remove this and have JS files clearly separated and identified.
- the integration allows `js-conf` and `js-exec` for execution before and after qTranslate respectively. No one uses `js-conf` it seems.
- the grain is too fine, this tends to multiply the number of executed scripts with one script per entry. When bundling an application we want rather the opposite, regrouping many parts in a single bundle. Though the script may be bigger this is usually better as it's in the cache and reduces the number of connections. Still we may have separate scripts per supported modules / plugins / themes but this could evolve.
-  instead of having `js-exec` deciding which script to load, an idea could be to use the keys of the pages instead and it should be up to the JS scripts to do the branching internally.

I don't think many people use the JS integration since it's been so cumbersome. The main plugins are now integrated as modules and the first concerned is ACF that we can handle directly (that was the idea of taking the plugins as modules internally, allowing easier refactoring). Perhaps I'll have a phase of deprecation but i'm tempted to break the `js-exec` config entry.

Any comments welcome.

picasso:
I used `js-exec` a lot.

I didn't really understand this idea:
> instead of having js-exec deciding which script to load, an idea could be to use the keys of the pages instead and it should be up to the JS scripts to do the branching internally.

could you explain?

herrvigg:
Great, it would be interesting to see how you use it.
If you look at every entry in `admin-config` you have for example this:
```
{
 "post": {  // <---- this is the key
      "pages": {
        "post.php": "",
        "post-new.php": ""
      },
      "anchors": ...,
      "forms":  ...,
      "js-exec": {
        "post-exec": {
          "src": "./dist/post-exec.js"
        }
      }
    },
  }
```

The main logic behind the JSON configurations is selectors / LSB placement / fields selectors + behavior / js code. The first key identifies each block, in this case it's called `post` and logically the JS script is called `post-exec.js`. 

[...]

PR #1009, opened 2021-04-18
Refactor 'js-exec' config entries with JS events

herrvigg:
Deprecate the `js-exec` in the JSON config and prevent embedded JS code in JSON, see #1000. This simplifies qTranslate, improves security and the integration framework.
Events are now triggered from the core module during JS execution. This also gives a much better control of the loading sequence. 

## Execution

New events named  `'qtxLoadAdmin:[page-key]'` are now triggered from the core module on `document`. The main `qtx` object is passed as parameter.

Example of handler: `$(document).on('qtxLoadAdmin:post', function( event, qtx ) { /*...*/ } );`.

The page keys come from the `admin-config` entry in the i18n-config.json. Only the active keys for the page currently visited are passed, they could be several keys corresponding to different page selectors.

There must be some `pages` or `post_type` selector making the current page specific for the event to be fired. Entries without specific selector are irrelevant since they would apply to all pages.

More detailed examples:
### post
```
// i18n-config.json
"admin-config": {
    "post": {   // <--- page key
      "pages": {
        "post.php": "",
        "post-new.php": ""
      },
    // "js-exec" --> obsolete, remove this entry
```
```
// JS handler for exec (ES5)
$(document).on('qtxLoadAdmin:post', function(event, qtx) {
   // do exec for this entry...
});
```

### nav-menus
```
// i18n-config.json
"admin-config": {
    "nav-menus": { // <--- page key
      "pages": {
        "nav-menus.php": "action=edit|menu=\\d+|^$"
      },
    // "js-exec" --> obsolete, remove this entry
```
```
// JS handler for exec (with ES6 arrow function)
$(document).on('qtxLoadAdmin:nav-menus', (event, qtx) => {
   // do exec for this entry...
});
```

## Loading scripts

Custom scripts should not be loaded by qTranslate anymore. The third-party plugins or themes should handle this, there could be some specificities that are not part of qTranslate and these scripts don't belong to qTranslate. They should be loaded in the footer. The only constraint is the dependency to `qtranslate-admin-main` to use the `qtx` API. 

Still we might have a default script to load to facilitate basic JS functionalities for simple plugins and themes.

## Front config

Since qTranslate has no JS script running on the front-end, nothing is done. Custom scripts should also be handled by third-party components.


herrvigg:
I'm not sure yet for the events if we should have:
- generic events with parameters:  `qtranslate_load_admin` + page_key (single) or page_keys (array)
- specific events: `qtranslate_load_admin_<page_key>`, e.g. `qtranslate_load_admin_post`, `qtranslate_load_admin_nav_menus`, ...
- maybe a combination of those

Issue #1018, opened 2021-05-09
Broken editor with single language and visual mode


herrvigg:
The editor content is corrupted when starting in visual mode and single language mode (advanced settings of QTX).
New lines disappear. The problem is related to the handling of `wpautop`.

It works when  starting in text mode and switching to visual mode.
The translatable class is missing in that case.

MK-RD:
QTX 3.10.0

For some reason I cannot reproduce this behavior. 
Classic Editor or Block Editor?

herrvigg:
Classic Editor, you have to start in Visual Mode when loading the post. The content was corrupted with new lines deleted.

Fixed in 3.10.1.

PR #1019, opened 2021-05-09
Remove obsolete wpautop hack for editor init 

herrvigg:
Many complications come from a legacy hack to deal with a `wpautop` filter related to the editor content.
The idea was to remove the filter `wp_richedit_pre` on PHP side when the editor was TinyMCE and to restore it on the JS side with the infamous "ugly trick" introduced in 3.2.9.8.

It turns out that `wp_richedit_pre` was deprecated already in WP 4.3.0... making the whole hack obsolete.
On the PHP side, it is now `format_for_editor` that applies to TinyMCE (visual mode), but `wpautop` is not called anymore.
Many transformations have been done in WP to handle `wpautop` on the JS side.

So all the legacy mess in QTX seem completely unnecessary now. Since the ML content is parsed on the JS side and `wpautop` is almost entirely handled by WP editor, all should become much simpler. The only thing QTX should care of is when it changes the editor content on a language switch. In that case we should just rely on the `wpautop` flag defined for the editor.

There might still be some special cases to handle for ACF but that were most likely broken already (for example #910).
The changes of this PR should be a first step towards a more reliable behavior.


herrvigg:
This also fixes the issue with single language mode #1018.
The editor hooks are not initialized for single language mode and the translatable class are added.
In contrast to the RAW mode, the single language mode is still performing some ML manipulations on the JS side, that are recomposed on the PHP side. So it is logic to display the translatable bars.

PR #1033, opened 2021-05-26
Fix 'name' schema for WebPages

pjrobertson:
Copying this over from my current local changes into master


herrvigg:
Many thanks!

PR #1039, opened 2021-06-09
Run a filter before outputting the hreflang link

reekrd:
This allows users to skip generating a hreflang link

### DIFF
--- a/qtranslate_frontend.php
+++ b/qtranslate_frontend.php
@@ -55,7 +55,9 @@ function qtranxf_wp_head() {
     $hreflangs['x-default'] = qtranxf_convertURL( '', $q_config['default_language'] );
 
     foreach ( $hreflangs as $hreflang => $href ) {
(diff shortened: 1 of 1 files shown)


herrvigg (commented):
I made something similar temporarily in master but removed it until the question of the locales is better solved. I'd prefer a unique filter fired once instead of firing for every language.

reekrd:
Is it a better approach to run it between the two loops?

herrvigg:
It could be done with an array indexed by `lang`, with `url` and `locale`. A filter could change/remove some items and QTX would only display what is left.

herrvigg:
Ah, i only see the new version now. Yes, something like that, but I'd still need to clarifiy if we need a separation between `lang` and `locale` for the `hreflang` key.

herrvigg:
Sorry for long delay but it's related to a wider topic with #668 so I have to keep it on hold until this is addressed, lower priority at the moment. Patching the code locally is the only way for now, but I know this is not great 😕

Issue #616, opened 2018-11-20
Language not working in widgets


hubosh:
I just started using qTranslate XT and I was hoping for a fix to this problem.
When using text in a widget I can't change the languages, it just copies the last data I entered and stores it in all language tabs.

herrvigg:
Can you give more details about this problem? I guess you are referring to switching tabs but i can't see in which context. The widgets in the sense "qtranslate Widget" are meant to change language on the front-end so there's no data to enter there. So i suppose it's an admin page but which one? A screenshot could also help.

herrvigg:
About the widget admin menu, the only translatable field is the title and this works fine for me.

![image](https://user-images.githubusercontent.com/19155375/48771333-6e68b800-ecc1-11e8-8017-386b3c241f73.png)

hubosh:
> Can you give more details about this problem? I guess you are referring to switching tabs but i can't see in which context. The widgets in the sense "qtranslate Widget" are meant to change language on the front-end so there's no data to enter there. So i suppose it's an admin page but which one? A screenshot could also help.

What I meant is when I use text in a widget, I can't translate it, not even the title.
Here's a screenshot
![screenshot 2](https://user-images.githubusercontent.com/24465260/48775789-2193da00-ecdf-11e8-8a0e-71005909ea60.png)

forexmd:
I also have problem with translating widgets, on old qtranslate-x and the same on qtranslate-xt , which i just installed. When i save some text on one language immediately appears the same text for the second language. Do you have any idea how to make it work normal. Website forex.md . Thanks

herrvigg:
OK now i see, i can reproduce the problem. The input and textarea fields don't have the blue marker, which means they are not considered as translatable but i guess they are supposed to be part of those. I tried quickly to update the json config but it didn't work. I'll see if i manage to fix it.

forexmd:
whatever it is worth, the problem is only on the latest version of WP. With previous version of WP, was no problem, i have always made changes in footer/widgets before and had no problems with that.

herrvigg:
Oh i thought it was a much older problem.
If you look into i18n-config.json the selectors are `input[id^='widget-'][id$='-title']` and `textarea[id^='widget-text-'][id$='-text']` but this does not hold anymore. I tried other selectors simply with `input.title` and `textarea.text` but it doesn't work. At least we know the `widget-right` selector works as the switching tabs are shown.
[...]

PR #1042, opened 2021-06-26
Enable language switch for text widget

herrvigg:
The text widget was broken since 2017 though the main issue was raised in 2018: #616.

It is a very complex feature to restore due to the way the widgets are updated in javascript with many dynamic behaviors and "internal" WP sync fields: https://make.wordpress.org/core/2017/05/23/addition-of-tinymce-to-the-text-widget/

There are several situations to handle:
- existing widget initialized with pre-loaded content 
- widget updated (save button)
- widget added with or without existing switching buttons on the page.

This major rework of the text widget in qTranslate-XT requires several changes in the JS API with new functions exposed:
- `attachContentHook`
- `attachEditorHook` (*)
- `setupLanguageSwitch`

(*)  `attachEditorHook` is renamed from `setEditorHooks` but normally third-party code should rather use `loadAdditionalTinyMceHooks()` for standard TinyMCE hooks.


Hr0bar:
Hi herrvigg, thanks again for this work, when can we expect this to be released?

herrvigg:
I was about to merge this but WP has released a new feature in WP 5.8 to manage the widgets... with Gutenberg.
Hopefully there's a new plugin "Classic Widgets" which disable this feature. It's very simple, by adding two filters, but I need to check this better and possibly do a similar workaround with Gutenberg using a single language mode.

herrvigg:
Released in 3.11.0.

PR #1049, opened 2021-07-29
update zh_CN (100%)

dallaslu:


herrvigg:
Thank you!

herrvigg:
hi @dallaslu 
just a short notice, I updated the localization file you gave me from the current code (qTranslate-XT 3.11.0).
The translations are now at 80% done. Since it was 100% from your version, you probably did it from a very old state of the code. There has not been many changes with new translation labels in the last months.
In any case, thank you very much for your update, it is already a good step ahead! 👏

PR #1059, opened 2021-08-30
update zh_CN (100%)

dallaslu:


herrvigg:
Wow, you are fast!
I realized there are a few strings we don't want in the PO files coming from Javascript, due to wrong legacy filters with the functions. In fact you recopied those values as they were, it's fine. I'll clean this up before merge but it will remain at 100%.

Issue #671, opened 2019-04-22
Include qtranslate slug into -XT


The3IC:
Now that the -XT work is progressing so well, I would like to request that the old qTranslate slug be integrated. At this time it works still well with -XT but to make sure that this very valuable functionality is maintained with -XT, it would be great if it was made part of -XT.

https://github.com/not-only-code/qtranslate-slug

herrvigg:
Sure it would be good to have, but it's not my priority for now.

As i mentioned in https://github.com/qtranslate/qtranslate-xt/issues/660, the problem is qTranslate slugs has not been updated for 3 years and it has a lot of open issues:
https://wordpress.org/support/plugin/qtranslate-slug/

The git repo is not fresh either and there are also many open issues without answer:
https://github.com/not-only-code/qtranslate-slug/issues

Even if it looks okay for some people, I will definitely not merge a plugin in that state. better to keep it separated at this point. It would be good to have the authors involved, but the first step would be to refresh the plugin even before talking about merging it here. What you (or anyone else interested in this) can do is start to contact them on their repos and tell we are reviving qTranslate(-X/XT) so they can also start to revive their stuff ;)

The other option is to have new persons dedicating their time on that. A lot of testing would be required and then it can goes in with a PR but i am quite reluctant to merge something risky without the full knowledge the original authors had so i would definitely prefer the first option.

gm-style:
Hi there, I'm actively testing your new qT-XT with some popular plugins (woocommerce, ACF and Yoast) and latest WP versions. Actually I installed the "old" qT Slug plugin and it seems working like a charm, even with WC categories. I didn't find any issues atm but I'll keep testing.

The3IC:
> Actually I installed the "old" qT Slug plugin and it seems working like a charm, even with WC categories. I didn't find any issues atm but I'll keep testing.

qT Slug (as qT itself) work amazingly well considering they have gone a couple of years without maintenance. This just speaks a lot about the good initial architecture (compare to the constant flow of WPML problems). However, there are a number of quite important error reported over in the qT Slug repo, that would deserve getting fixed. In my view the worst problem is that the Slug breaks subpages. There are also problems with taxonomies (mainly "follow-on problems such as problems with SEO etc) and personally I'm sorely missing the capability to add language specific URLs in the Menus.

In general all (reasonably) successful language extensions do support slug translation so therefore I think is quite important for XT to have that functionality to get seriously back in the game!

[...]

PR #1060, opened 2021-08-31
Raw merge of qtranslate-slug 1.1.8 as qtranslate-xt module

spleen1981:
Initial merge with minimal modifications to both qtranslate-xt core and qtranslate-slug to incorporate the latter as a module.
Module can be enabled through a checkbox in qtranslate-xt advanced settings.
Heavy code cleanup and harmonization yet to be performed.
Settings page also to be incorporated in qtranslate-xt page


herrvigg:
Need more time to check and polish, but it's in the plan to integrate this. It will be a major release.

modules/qtx_modules_handler.php
```diff
@@ -89,8 +102,13 @@ public static function get_modules_defs() {
                 'name'         => 'Yoast SEO',
                 'plugin'       => 'wordpress-seo/wp-seo.php',
                 'incompatible' => 'wp-seo-qtranslate-x/wordpress-seo-qtranslate-x.php'
+            ),
+            array(
+                'id'           => 'slugs',
+                'name'         => 'Slugs',
+                'plugin'       => true,
```
herrvigg, line 109: We should rather enable that if the user checks "enable experimental slugs", so the other way around.
But I see why it is easier like this for now. The module has still the legacy code of the plugin, so the init sequence should be refactored.

qtranslate_core.php
```diff
@@ -61,11 +61,6 @@ function qtranxf_init_language() {
         }
     }
 
-    // TODO fix qtranslate-slug still using 'original_url' field and remove it from here, this has no sense!
-    if ( defined( 'QTS_VERSION' ) ) {
-        $url_info['original_url'] = $_SERVER['REQUEST_URI'];
-    }
-
```
herrvigg, line 68: Nice, good to get rid of this!

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -0,0 +1,2700 @@
...
+        return $this->old_data;
+    }
+
+    function qtranslate_updated_settings() {
+        global $q_config;
+        $options_modules = get_option( 'qtranslate_modules', array() );
+
+        if ( $q_config['slugs_enabled'] ) {
+            $this->install();
+            $options_modules['slugs'] = 1;
```
herrvigg, line 324: Use define QTX_MODULE_STATUS_ACTIVE

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -0,0 +1,2700 @@
...
+        } else {
+            $this->deactivate();
+            $options_modules['slugs'] = 2;
```
herrvigg, line 327: Use define QTX_MODULE_STATUS_INACTIVE

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -0,0 +1,2700 @@
...
+
+        return (
+            version_compare( $wp_version, "4.0", "<" ) ||
+            ! ( defined( 'QTRANSLATE_FILE' ) ||
+                ( is_plugin_active( 'qtranslate/qtranslate.php' ) ||
+                  is_plugin_active( 'qtranslate/qtranslate-x.php' ) ||
+                  is_plugin_active( 'mqtranslate/mqtranslate.php' ) ||
+                  is_plugin_active( 'qtranslate-x/qtranslate.php' ) ) )
+        );
+    }
```
herrvigg, line 100: All of that becomes obsolete. Can be cleaned up later.

modules/qtx_modules_handler.php
```diff
@@ -29,6 +32,16 @@ public static function load_modules_enabled() {
         }
     }
 
+    public static function update_manual_enabled_modules() {
+        global $q_config;
+        $options_modules = get_option( 'qtranslate_modules', array() );
+
+        if ( $q_config['slugs_enabled'] && $options_modules['slugs'] != 1 ) {
+            $options_modules['slugs'] = 1;
```
herrvigg, line 40: Use define QTX_MODULE_STATUS_ACTIVE

admin/qtx_admin_settings.php
```diff
@@ -693,6 +705,17 @@ private function add_advanced_section() {
...
+                <th scope="row"><?php _e( 'Slugs translation', 'qtranslate' ) ?></th>
+                <td>
+                    <label for="slugs_enabled">
+                        <input type="checkbox" name="slugs_enabled"
+                               id="slugs_enabled"
+                               value="1"<?php checked( $q_config['slugs_enabled'] ) ?>/>&nbsp;<?php _e( 'Enable slugs translation.', 'qtranslate' ) ?>
+                    </label>
+                    <p class="qtranxs-notes"><?php echo __( 'This will activate the slug translation module (experimental feature). ', 'qtranslate' ) ?></p>
+                </td>
+            </tr>
```
herrvigg, line 718: I think this can move to the integration panel, because it is directly related.
This functionality should certainly exist for any module in the future.

herrvigg (commented):
Nice job!! Looks good overall!
A few remarks with a few changes to come before merge.
We should transition fully from plugin to module. It is still in an intermediate state. But the idea is to merge and continue refactoring this on master.

herrvigg:
For the last updates and post-merge cleanup see #671.

bagaweb:
Hello maybe this is not the correct place to post this error, let me know if I should open a new thread... 

Anyway... I noticed an issue with Russian language, if the slugs are in Cyrillic (like the titles), front-end links in Russian language give 404 error.

If I rewrite the slugs in Latin characters (for example apartamenty instead of апартаменты), there are no issues.

Anyone else noticed this bug?

spleen1981:
Based on original plugin [README](https://github.com/qtranslate/qtranslate-xt/tree/master/modules/slugs#todo), this seems to be a known issue.

bagaweb:
> Based on original plugin [README](https://github.com/qtranslate/qtranslate-xt/tree/master/modules/slugs#todo), this seems to be a known issue.

Ok but this issue was about taxonomies and custom post types, not about pages.

On an another project which is using QTX & QTS 1.18, pages work fine even with Cyrillic slugs.

bagaweb:
> I noticed an issue with Russian language, if the slugs are in Cyrillic (like the titles), front-end links in Russian language give 404 error.
> 
> If I rewrite the slugs in Latin characters (for example apartamenty instead of апартаменты), there are no issues.
> 
> Anyone else noticed this bug?



> @bagaweb opening a new issue would be better for proper follow-up and visibility.

BTW I opened a new issue here: https://github.com/qtranslate/qtranslate-xt/issues/1156

Issue #256, opened 2018-06-22
Menu items without label are removed from the menu


herrvigg:
<a href="https://github.com/cdauth"><img src="https://avatars3.githubusercontent.com/u/1801792?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cdauth](https://github.com/cdauth)**
_Friday Sep 11, 2015 at 13:51 GMT_
_Originally opened as https://github.com/qTranslate-Team/qtranslate-x/issues/256_

----

Menu items with an empty label are removed from the menu. For example, when adding a page that only has an English translation (and thus its Italian title is empty), in the Italian version of the menu this page will be missing. This behaviour is unexpected, as there is a setting “Hide Content which is not available for the selected language”, which is not respected for menus.

The expected behaviour would be that either the “Hide Content” setting is respected for menu items with an empty label, or even better to have a separate global or per-menu setting for this specific issue.

herrvigg:
<a href="https://github.com/johnclause"><img src="https://avatars2.githubusercontent.com/u/5491361?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [johnclause](https://github.com/johnclause)**
_Friday Sep 11, 2015 at 17:41 GMT_

----

Correct, "Hide Content .." is for content of posts and pages only, not menu, not widgets etc. Empty menu labels allow to manipulate menus in a more flexible way. How would you show menu item with empty label anyway? What is it exactly you are trying to achieve?

herrvigg:
<a href="https://github.com/cdauth"><img src="https://avatars3.githubusercontent.com/u/1801792?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cdauth](https://github.com/cdauth)**
_Friday Sep 11, 2015 at 18:12 GMT_

----

I would display it in the same way that any other content is displayed, if there is no translation, fall back to the default language.

I have a website where a few pages are translated into a lot of languages, and a lot of pages are only in English. What I'm trying to achieve is that all menu items appear in all languages, even if the page is not translated into that language. With the way that qtranslate-x currently works, I would have to copy the English page title / menu item label into every language where a translation doesn't exist.

herrvigg:
<a href="https://github.com/johnclause"><img src="https://avatars2.githubusercontent.com/u/5491361?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [johnclause](https://github.com/johnclause)**
[...]

PR #1063, opened 2021-09-09
Add setting to disable hiding untranslated menu items

cdauth:
This pull request adds a setting to the admin page where users can configure whether menu items should be hidden when their label (with a fallback to the page title) is not translated into the current language.

The default behaviour in QTranslate was to show such items. The default behaviour in QTranslate-X and QTranslate-XT has been to hide such items. Because of this, the default setting is to hide them.

The issue has been discussed in more detail in #256.

It's my first time contributing to QTranslate-XT, so I'm not sure if I did everything right. In particular, I'm not sure whether/how I should recreate the qtranslate.pot file.


herrvigg:
Thank you! The changes look good. I need to read all the story to understand the topic better.

cdauth:
This is a common use case in smaller non-commercial collectives or associations: The website is written in one main language (for transnational collectives this is usually English), but a small number of important pages (such as the front page) are translated into other languages, so that people who don't speak the main language can still get a rough idea what the website is about. There are not enough resources to translate all content into all languages.
However, many of the visitors are bilingual and speak the main language in addition to their preferred language. Because the website will be shown to them in their preferred language but many pages are not translated into it, those pages should still show up in the menu, so that the visitors can see that this content exists and they can choose to open it if they speak that language as well (or want to run it through a translator).

herrvigg:
Thank you for the summary.

As the previous maintainer I also tend to limit the options as much as possible but yes, in that case it makes sense to give this possibility and this seems as the simplest way to do so. If there are many menus and languages we should not impose all the menus to be translated., which might take time and resources. I won't discuss the reasons nor the design for having too many menus, those choices are up to the responsible of the website.

A few questions though.

1) I also wondered about the option to show the language prefix, You disabled it but it seems we "waste" the option in that case. I don't see any case where you'd want to see this in the menu, it's way too invasive. I'd rather see a flag or a short abbreviation but that's more work. As a baseline I'd rather say the prefix should be implicitly hidden for menus displayed for different languages. The prefix option should only be for posts or other items but not for menus.

2) There's a piece of code in the menus I need to understand better. We might have to rewrite this differently:
https://github.com/qtranslate/qtranslate-xt/blob/4930a23b3442d70189b59b900947ea4ce4d3ac45/qtranslate_frontend.php#L98-L99

3) Finally in #256 there was a mention about taxonomy. Maybe that should be handled similarly, or not.

cdauth:
1. Do I understand you right, that you think that the option “Show displayed language prefix when field content is not available for the selected language.” should have no effect for menu items?
2. I'm not sure how that's related to this issue. How I understand that code is that it is a way to fetch the menu item label in some edge cases. It doesn't seem to me that it has any effect on the issue that this pull request is addressing.
3. I have no experience with taxonomies in Wordpress and how they are handled by QTranslate. Does anything need to be adapted in this pull request?

herrvigg:
1. Yes that's what I mean. But it turns out to be more complicated than expected. We can't change that without changing the API for several functions. Let's keep it out of scope for now.
2. Yes this case is tricky and I don't even understand it well... But there's at least something to clarify with the comment in the code which is very confusing.
3. The taxonomy was mentioned in #256 so maybe we should combine this to have a unique option. Let's dig a bit more before merge.

herrvigg:
I don't forget this, it's almost ready to merge but need a bit more time to check the third point. 
I will push a new release without it, I prefer not to rush it to avoid back-and-forth with the user options.

herrvigg (approved):

herrvigg:
@cdauth Sorry this has been so long... I have updated the options naming and description but I keep it as a separate option item. About taxonomy it's another topic or a specific use case, so let's merge the new feature. The use case makes sense as your described it, it can be meaningful for many and it's possible to continue building on top of this.

Issue #1077, opened 2021-11-01
Fixes for Yoast SEO module


Komarovski:
Hello everyone,
I've made some frontend fixes for Yoast SEO module that prevent outputting wrong language data in breadcrumbs from indexable tables and also added filters for schema, opengraph and next|prev links.

Tested on:
Wordpress version: 5.8.1
qTranslate XT version: 3.11.0
Yoast SEO version: 17.4

Here is the source-code for wp-seo-front.php, please review this fixes - maybe some refactoring will needed:
[wp-seo-front.zip](https://github.com/qtranslate/qtranslate-xt/files/7453960/wp-seo-front.zip)

Komarovski:
All new lines of code contains "//NEW" comments.

herrvigg:
Thanks! I'll take a look.

herrvigg:
Created #1079. I have a few questions, but let's discuss them in the Pull Request. We can close this one.

PR #1079, opened 2021-11-14
Fix yoast breadcrumbs front

herrvigg:
From #1077 by @Komarovski
> Hello everyone, I've made some frontend fixes for Yoast SEO module that prevent outputting wrong language data in breadcrumbs from indexable tables and also added filters for schema, opengraph and next|prev links.
> 
> Tested on: Wordpress version: 5.8.1 qTranslate XT version: 3.11.0 Yoast SEO version: 17.4

### DIFF
--- a/modules/wp-seo/wp-seo-front.php
+++ b/modules/wp-seo/wp-seo-front.php
@@ -4,7 +4,19 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 function qtranxf_wpseo_add_filters_front() {
+    // Use indexation only in backend on "publish/update" events to save all languages data in indexable tables.
+    // If you allow indexation on the frontend - then indexable table will save data of the first visited page, all other languages data will be missed/filtered/gone/didn't saved.
+    // After all this fixes everyone need to install "Yoast Test Helper", go to "Instruments -> Yoast Test" and reset "Indexables & migrations table", then go to "SEO -> Instruments" and start SEO data optimization.
+    // This filters "wpseo_should_save_indexable", "wpseo_indexing_data" - right now works only on frontend, so using of "!is_admin()" is pointless, but maybe in future this filters will work on backend, that's why I have leaved "!is_admin()" below.
+    // Unfortunately I didn't found how to totally disable indexable feature, Yoast documentation is very poor to quickly understand how everything works.
+    // Also be advised that after Yoast have added indexables - now you also need to provide breadcrumb titles and its translations even if they are same as in title. If you wouldn't provide breadcrumb title & translation - Yoast will only index current language title.
+    if ( ! is_admin() ) {
+        add_filter( 'wpseo_should_save_indexable', '__return_false' );
+        add_filter( 'wpseo_indexing_data', '__return_false' );
+    }
+
     add_filter( 'wpseo_canonical', 'qtranxf_checkCanonical', 10, 2 );
+    add_filter( 'wpseo_opengraph_url', 'qtranxf_checkCanonical', 10, 2 );
 
     # For reference: https://developer.yoast.com/customization/apis/metadata-api/
     $use_filters = array(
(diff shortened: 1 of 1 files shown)


modules/wp-seo/wp-seo-front.php
```diff
@@ -4,7 +4,19 @@
...
+    // If you allow indexation on the frontend - then indexable table will save data of the first visited page, all other languages data will be missed/filtered/gone/didn't saved.
+    // After all this fixes everyone need to install "Yoast Test Helper", go to "Instruments -> Yoast Test" and reset "Indexables & migrations table", then go to "SEO -> Instruments" and start SEO data optimization.
+    // This filters "wpseo_should_save_indexable", "wpseo_indexing_data" - right now works only on frontend, so using of "!is_admin()" is pointless, but maybe in future this filters will work on backend, that's why I have leaved "!is_admin()" below.
+    // Unfortunately I didn't found how to totally disable indexable feature, Yoast documentation is very poor to quickly understand how everything works.
+    // Also be advised that after Yoast have added indexables - now you also need to provide breadcrumb titles and its translations even if they are same as in title. If you wouldn't provide breadcrumb title & translation - Yoast will only index current language title.
+    if ( ! is_admin() ) {
```
herrvigg, line 13: If it's admin it should rather go to `wp-seo-admin.php`.

modules/wp-seo/wp-seo-front.php
```diff
@@ -38,9 +50,87 @@ function qtranxf_wpseo_webpage_schema( $piece, $context ) {
     add_filter( 'wpseo_breadcrumb_single_link_info', 'qtranxf_wpseo_breadcrumbs_link', 10, 3 );
     function qtranxf_wpseo_breadcrumbs_link( $link_info, $index, $crumbs ) {
         $link_info['text'] = qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage( $link_info['text'] );
+        $link_info['url']  = qtranxf_convertURL( $link_info['url'], $lang );
```
herrvigg, line 53: The `$lang` is not set. This result in using the current language `$q_config['language']`. Is that the intention?
Same below for all filters.

modules/wp-seo/wp-seo-front.php
```diff
@@ -38,9 +50,87 @@ function qtranxf_wpseo_webpage_schema( $piece, $context ) {
...
+    function qtranxf_wpseo_schema_webpage( $data ) {
+        $data['@id'] = qtranxf_convertURL( $data['@id'], $lang );
+        $data['url'] = qtranxf_convertURL( $data['url'], $lang );
+        if ( $data['isPartOf']['@id'] ) {
+            $data['isPartOf']['@id'] = qtranxf_convertURL( $data['isPartOf']['@id'], $lang );
+        }
+        if ( $data['primaryImageOfPage']['@id'] ) {
+            $data['primaryImageOfPage']['@id'] = qtranxf_convertURL( $data['primaryImageOfPage']['@id'], $lang );
+        }
+        //if($data['author']['@id']){$data['author']['@id'] = qtranxf_convertURL($data['author']['@id'], $lang);} //Not sure is it required to filter or not???
```
herrvigg, line 92: Probably needed for consistency.

modules/wp-seo/wp-seo-front.php
```diff
@@ -38,9 +50,87 @@ function qtranxf_wpseo_webpage_schema( $piece, $context ) {
...
+
+        return $data;
+    }
+
+    add_filter( 'wpseo_schema_imageobject', 'qtranxf_wpseo_schema_imageobject' );
+
+    function qtranxf_wpseo_schema_webpage( $data ) {
+        $data['@id'] = qtranxf_convertURL( $data['@id'], $lang );
+        $data['url'] = qtranxf_convertURL( $data['url'], $lang );
+        if ( $data['isPartOf']['@id'] ) {
```
herrvigg, line 86: Better test with `isset` or `! isempty`. Do you know what is needed? I suppose `isset`, but not sure.

herrvigg (commented):
A few questions

modules/wp-seo/wp-seo-front.php
Komarovski: Yes, the intention to use current language.

modules/wp-seo/wp-seo-front.php
Komarovski: I've commented this line because I didn't understand how Yoast plugin output schema in frontend.

For example for webpage schema - items like "$data['@id']" and "$data['url']" have wrong urls without using a filter "qtranxf_convertURL". And for item "$data['author']['@id']" url is correct and points to the current language.

Same situation for all other types of schema - they all work correct except those that was filtered with "qtranxf_convertURL" function in "wp-seo-front.php".

Thoughts:
Looks like when Yoast plugin outputs schema in some cases they use URLs that is placed in "indexables" table - and that's why URLs are wrong in different languages and need to be filtered. And in other cases their plugin get URLs using default wordpress tables.

modules/wp-seo/wp-seo-front.php
Komarovski: Yes, better to test with `isset`.

modules/wp-seo/wp-seo-front.php
Komarovski: Oh, sorry, didn't mention that the filename called "wp-seo-FRONT" :)
In that case you can delete condition `if ( ! is_admin() )`

modules/wp-seo/wp-seo-front.php
herrvigg: OK I leave it commented for now.

modules/wp-seo/wp-seo-front.php
herrvigg: Done.

modules/wp-seo/wp-seo-front.php
herrvigg: Done.

modules/wp-seo/wp-seo-front.php
herrvigg: Done.

herrvigg:
Push a few fixes. I don't use these features myself so I trust you. I think we can merge this.

herrvigg:
Merged in master so more people can test. If you spot new errors let us know.

herrvigg:
Perhaps we should keep your comments somewhere in the wiki, the first part fort he indexation might be useful for the users:
> Use indexation only in backend on "publish/update" events to save all languages data in indexable tables.
> If you allow indexation on the frontend - then indexable table will save data of the first visited page, all other languages data will be missed/filtered/gone/didn't saved.
> After all this fixes everyone need to install "Yoast Test Helper", go to "Instruments -> Yoast Test" and reset "Indexables & migrations table", then go to "SEO -> Instruments" and start SEO data optimization.
> This filters "wpseo_should_save_indexable", "wpseo_indexing_data" - right now works only on frontend, so using of "!is_admin()" is pointless, but maybe in future this filters will work on backend, that's why I have leaved "!is_admin()" below.
> Unfortunately I didn't found how to totally disable indexable feature, Yoast documentation is very poor to quickly understand how everything works.
> Also be advised that after Yoast have added indexables - now you also need to provide breadcrumb titles and its translations even if they are same as in title. If you wouldn't provide breadcrumb title & translation - Yoast will only index current language title.

Komarovski:
> > After all this fixes everyone need to install "Yoast Test Helper",
> 
> @Komarovski what is the impact if we don't do so?

If don't do so - Yoast SEO will use old data from indexables table that could not have all translations in titles, descriptions, breadcrumbs.
Other way to update indexables table so it could store all translated content without using Yoast Test Helper - is manually update all posts, pages, custom post types items, categories, tags, custom taxonomy terms, author pages.

After updating to qTranslate-XT 3.11.1 version - this operation needs to be done one time.

PR #1081, opened 2021-11-21
Fix localization of ACF field labels

herrvigg:
~Localize strings of the ACF module in domain qtranslate (previously in acf).~
~No i18n translations were found in the legacy acf-qtranslate project.~

qTranslate should not add new strings to `acf`.


herrvigg:
Re-use localization strings available in the `acf` domain. No need to add extra work for the translators in qTranslate.

Issue #1086, opened 2021-12-01
Warnings with Yoast breadcrumbs front


herrvigg:
Issues following #1079 
```
Warning: Attempt to read property "target" on array
/wp-content/plugins/qtranslate-xt/modules/wp-seo/wp-seo-front.php in qtranxf_wpseo_schema_webpage at line 106
 $data['potentialAction'][0]['target'] = [ qtranxf_convertURL( $data['potentialAction'][0]->target, $lang ) ];
```

```
Warning: Undefined array key "url"
/wp-content/plugins/qtranslate-xt/modules/wp-seo/wp-seo-front.php in qtranxf_wpseo_breadcrumbs_link at line 50
$link_info['url']  = qtranxf_convertURL( $link_info['url'], $q_config['language'] );
```

herrvigg:
Will send a fix tomorrow. This has exploded my quotas on sentry...

herrvigg:
Released in patch 3.11.2 only to fix this issue, to avoid saturating the monitoring tools.

herrvigg:
Regression in the last patch... sorry about that! Hard to test those filters -.-
Fix sent by @Komarovski, thanks!

herrvigg:
New fix released in 3.11.3.
Small typo in the commit title, the regression came from 3.11.2 (not 3.11.1).
The changelog and release notes in 3.11.3 are correct.

These fixes are mostly to avoid saturating monitoring tools, should have no impact apart from the SEO data with Yoast.

herrvigg:
@Komarovski guess what... this is not over yet!
New problem: `$data['potentialAction'][0]['target']` is an array, which makes `qtranxf_convertURL` to fail...

Their schema is not documented very well but the example shows this is an array:
https://developer.yoast.com/features/schema/pieces/webpage
I don't even understand why they are creating so many complications with a crazy number of fields. All these repetitions look really insane to me... it doesn't make any sense at all.

Anyway, a new fix is needed.

herrvigg:
New fix released in 3.11.4...

I'm pretty sure there are many other schema fields that should be translated. But that is less problematic.

Now I start to have some doubts about the `@id` and URLs expecting **canonical** URLs... I'm not entirely sure the current language should be used to convert the URLs. What is the impact of the browser detection? But that's a new topic and I'll leave it to the Yoast users, it can be discussed separately.

PR #1087, opened 2021-12-03
Update wp-seo-front.php

Komarovski:
Add "isset" check on 105 line.


herrvigg:
I pushed a more complete fix for #1086, see https://github.com/qtranslate/qtranslate-xt/commit/a7168a0c8959442b75f4874ba8ed4871ec88a25a.

I will release this early as a patch, would be good if you could try it.

PR #1088, opened 2021-12-04
Remove `hreflang` from `a` tag in widget

Komarovski:
There's no need to set hreflang in "a" tag, since we have "link" tags with hreflang in "head" of the website.

I have digged all around the network and didn't find any information about using hreflang attribute inside of "a" tags.

### DIFF
--- a/qtranslate_widget.php
+++ b/qtranslate_widget.php
@@ -246,8 +246,6 @@ function qtranxf_generateLanguageSelectCode( $args = array(), $id = '' ) {
                         $classes[] = 'active';
                     }
                     echo '<li class="' . implode( ' ', $classes ) . '"><a href="' . qtranxf_convertURL( $url, $language, false, true ) . '"';
-                    // set hreflang
-                    echo ' hreflang="' . $language . '"';
                     echo ' title="' . $alt . '"';
                     if ( $type == 'image' ) {
                         echo ' class="qtranxs_image qtranxs_image_' . $language . '"';


herrvigg:
It's true that we mostly count on the `link` tag for this but I'm not sure about the `a` tag. I didn't dig much, this page might not be the best reference, but here `hreflang` is mentioned in the `a` tag: https://www.w3schools.com/tags/att_a_hreflang.asp. If we are really sure it's not needed then yes we can remove it.

Komarovski:
> It's true that we mostly count on the `link` tag for this but I'm not sure about the `a` tag. I didn't dig much, this page might not be the best reference, but here `hreflang` is mentioned in the `a` tag: https://www.w3schools.com/tags/att_a_hreflang.asp. If we are really sure it's not needed then yes we can remove it.

Sorry, yep, I have also found information (not official), that hreflang can be used in "a" tags. But here is another issue that we can fix by removing hreflang from "a" tags. Here is the situation:

For example, we have website with two languages "English (default)" and "Russian", plus we have "Pre-Path Mode" and "hide_default_language" is enabled. The output for "link" tags will look like:
```
<link hreflang="en" href="https://example.com/" rel="alternate">
<link hreflang="ru" href="https://example.com/ru/" rel="alternate">
```
Which is right, but in widget we will have another link for default language:
```
<a href="https://example.com/en/" hreflang="en" ***><span>English</span></a>
<a href="https://example.com/ru/" hreflang="ru" ***><span>Russian</span></a>
```
So "link" and "a" tags will point to different URLs "https://example.com/" and "https://example.com/en/" for same page.

herrvigg (approved):
OK let's remove this from the `a` tag.

PR #1089, opened 2021-12-04
Update qtranslate_core.php

Komarovski:
Use 301 redirect instead of 302 when switchng to default language with "hide_default_language" is checked. Good for SEO.

### DIFF
--- a/qtranslate_core.php
+++ b/qtranslate_core.php
@@ -88,7 +88,7 @@ function qtranxf_init_language() {
              */
             $target = apply_filters( 'qtranslate_language_detect_redirect', $url_lang, $url_orig, $url_info );
             if ( $target !== false && $target != $url_orig ) {
-                wp_redirect( $target );
+                wp_redirect( $target, 301 );
                 nocache_headers(); // prevent browser from caching redirection
                 exit();
             } else {


herrvigg:
I don't think it should be 301 since it could depend on the browser settings. In that case 302 is more appropriate for changing language redirections. If it were a fixed language from URL parameters then 301 could make sense. But this distinction is not made (yet).

Also, there's another `wp_redirect` in the same file.

Komarovski:
> I don't think it should be 301 in all cases since it could depend on the browser settings. In that case 302 is more appropriate because it's not going to the same target. If it were a fixed language from URL parameters then 301 could make sense. But this distinction is not made (yet).
> 
> Also, there's another `wp_redirect` in the same file.

Oh, yes, that makes sense now, I think we can close this request for now and meanwhile I will try to dig around the fixing of an language from URL parameters.

Sorry that I'm pulling so much requests and issues last time, it's just I'm trying to fix all this issues for my production website and in parallel publish them to github.

herrvigg:
No worries, it's great to see new pull requests. Perfect to separate them in different topics. I prefer to have several small PRs than a big one mixing a lot of unrelated changes.

herrvigg:
Closing this for now. we keep 302 that covers better the general case imo.

PR #1090, opened 2021-12-06
Add Yoast filters for organization schema and publisher

Komarovski:
Fix - we need to check for "$link_info" not the "$data"


herrvigg:
Ooops, my bad!! -.-
Good catch. I need something to test these filters earlier, it's not regular callbacks.
I'll have to send a new release again.

herrvigg:
I separated the commits to better track the history, but the fix is also for you credit ;)

PR #1095, opened 2022-01-05
Fixed an issue when trying to translate a GF choice text value

Juje:
This fixes a notice that is being triggered by updating a Gravity Form choice as an array even though it's a class.
And it will also make sure if it's an object or not (so that legacy versions will keep working).
Before this change on certain PHP versions the following notice will be triggered:

```
Indirect modification of overloaded element of GF_Field_Checkbox has no effect
```


herrvigg:
Looks good, thanks! Merged.

PR #1102, opened 2022-02-16
QTS: Use $post object instead of $_POST array

sviluppomania:
Use $post->post_title instead of $_POST['post_title']

WordPress 5.0.3/PHP-7.3.1


herrvigg:
I guess it's a good change. Have you seen any problem related to this?

herrvigg:
OK I made a quick test and $post contains raw ML values in that case.

PR #1107, opened 2022-02-19
slugs: fix for checkbox not enabling module on first save

spleen1981:


herrvigg (approved):
Update the state on save rather than load, it makes sense! Nice!

Issue #1097, opened 2022-01-12
full support of LSB in Gutenberg


picasso:
I implemented the full support of LSB in __Gutenberg__.
Since the solution has led to a significant amount of code, I put it into a separate plugin.
This plugin needs to be installed along with the `qTranslate-XT` and then you can customize it.
I think it can be a starting point for supporting Gutenberg. You can try the plugin from [my GitHub page](https://github.com/picasso/zu-translate).

For those who want to participate in the improvement/development of this solution, I created a note in which I briefly described the algorithm that was used to implement it. All explanations in the `GUTENBERG.md` file.

There are still a few problems that were not fully resolved, but I work on it. 
Also, improvements are possible, but it requires support from the main plugin - that is, changes in __qTranslate-XT__ code.
But I think it can be discussed after the main idea is accepted (and if accepted).

If the solution is suitable and will be quite stable and convenient, then in the future it will be possible to integrate it in __qTranslate-XT__.

herrvigg:
Wow, this looks big! I won't have much time to test it properly, there are other ongoing topics but we'll come back to this for sure. Meanwhile other users can also try it.

picasso:
I added an option to control `unsaved changes` after language switching. Also found a way to avoid a few __hacks__ and use the standard WP API. I will implement these ideas and then I will stop. For my projects, it's already enough, but it's still unclear if anyone else needs it.

Hr0bar:
Incredible !! If this gets core qTranslateXT we may actually keep using qTranslateXT forever!

Thanks for this.

picasso:
Implemented a converter from the `classic blocks` to __Gutenberg__ blocks. The converter is quite simple and processes only `headings`, `paragraphs` and `shortcodes`. In my opinion this is what is needed for most pages. Tested on my sites, but I cannot say that testing was complete and deep.
In any case, when converting a copy of page is created, so there will be no irreparable damage.
With this release the work on the support for Gutenberg, I consider the completed.

vellotis:
I'm not sure if I'm doing anything wrong but the two plugin combination didn't work for me.

## Steps
### 1. Use Docker to spin up Wordpress

> **`docker-compose.yml`**
> ```yml
> services:
>   db:
>     image: mariadb:10.4.22
>     volumes:
>       - db_data:/var/lib/mysql
>     restart: always
>     environment:
>       MYSQL_ROOT_PASSWORD: somewordpress
[...]

PR #1112, opened 2022-02-21
short circuit "QTX_Admin_Gutenberg"

picasso:
apply filter to short circuit the Admin handler for Gutenberg.
It is very necessary if you implement your processing for the Gutenberg Block Editor and  the related REST API.
Otherwise, you have to remove each filter and the action separately and sometimes with hacks...

### DIFF
--- a/admin/qtx_admin_gutenberg.php
+++ b/admin/qtx_admin_gutenberg.php
@@ -11,7 +11,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * Class QTX_Admin_Gutenberg
  *
- * Manages the Gutenberg block editor with the related REST API
+ * Manages the Gutenberg block editor with the related REST API.
+ * Limitation: only the single language mode is supported.
  */
 class QTX_Admin_Gutenberg {
     /**
@@ -26,11 +27,13 @@ class QTX_Admin_Gutenberg {
      * Register the REST filters
      */
     public function rest_api_init() {
-        // filter to short circuit the Gutenberg related REST API
-        $pre = apply_filters('qtranslate_gutenberg', false);
-        if($pre !== false) return;
-
         global $q_config;
+
+        // Filter to allow qTranslate-XT to manage the block editor (single language mode)
+        $admin_block_editor = apply_filters( 'qtranslate_admin_block_editor', true );
+        if ( ! $admin_block_editor ) {
+            return;
+        }
 
         $post_types = get_post_types( array( 'show_in_rest' => true ) );
         foreach ( $post_types as $post_type ) {
@@ -164,9 +167,11 @@ class QTX_Admin_Gutenberg {
      * Enqueue the JS script
      */
     public function enqueue_block_editor_assets() {
-        // filter to short circuit the Gutenberg JS script
-        $pre = apply_filters('qtranslate_gutenberg', false);
-        if($pre !== false) return;
+        // Filter to allow qTranslate-XT to manage the block editor (single language mode)
+        $admin_block_editor = apply_filters( 'qtranslate_admin_block_editor', true );
+        if ( ! $admin_block_editor ) {
+            return;
+        }
 
         wp_register_script(
             'qtx-gutenberg',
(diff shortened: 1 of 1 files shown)


sviluppomania (approved):

herrvigg (changes requested):
If you need to disable all the QTX functionalities for Gutenberg, there's a better way.
The instance should not be created so the filter call should be done in `qtx_admin_gutenberg.php` before requiring the file.

picasso:
> The instance should not be created so the filter call should be done in qtx_admin_gutenberg.php before requiring the file.

I wanted to do so. But who guarantees that at the time of creating an instance this filter will be already added? There is no guarantee that even a plugin that wants to do it is already loaded (QTX can be loaded before it). That is why I did it that way.

herrvigg:
Right, I see. Currently the instance is created from `qtranxf_init_language()` that is triggered on `plugins_loaded` with a high priority (2). So at this point we normally know which plugin is loaded technically, but they may not have set their filters yet.

One solution could be to delay the creation of this object, because it's only dependent on `rest_api_init` and `enqueue_block_editor_assets` which are fired much later, after `init` and even after `wp_loaded`. But that could compromise future evolutions if we'd need to setup new hooks earlier.

Another way could be to have a user option to disable the whole Gutenberg functionality. But I'm not sure it's better.

herrvigg:
A third possibility would be to give the possibility to _delete the object_ later instead of keeping it alive but having to check an external state in every function. Perhaps that would be a cleaner way to remove all filters. The cleanup would be done in the destructor.

picasso:
> Another way could be to have a user option to disable the whole Gutenberg functionality. But I'm not sure it's better.

yes, but hardly users will disable it. It is necessary to developers and third-party plugins. Therefore, my option is simple and will work with any future evolutions. Although not so elegant.

herrvigg:
Why do you need to disable these filters? Is it related to #1097 and https://github.com/picasso/zu-translate?

picasso:
> Why do you need to disable these filters? Is it related to https://github.com/qtranslate/qtranslate-xt/issues/1097

yes, I need to re-implement the support for Gutenberg and the current implementation interferes to me.

picasso:
> A third possibility would be to give the possibility to delete the object later
The object is so small that it makes no sense to optimize this process too much. Let it stay in memory, this is a small loss (IMHO)

herrvigg:
The idea of the object was to facilitate the deactivation of the filters (in the destructor), but this doesn't matter much.
I'm still thinking how to avoid the filter, but it looks like the most reliable solution. "Gutenberg" was more of a code name for the Block editor project.  We could rename the filter `qtranslate_block_editor`. I can send a patch.

picasso:
> We could rename the filter qtranslate_block_editor. I can send a patch.

Wait a bit. I began to make changes to my plugin because I have to return the functionality of the previous `QTranslate` to work with `WooCommerce`, as they do not support the "Gutenberg". And therefore it is necessary that part of the pages under the `post.php` worked as before. Therefore, we definitely cannot delete the object completely and can disable it or enable at the moment when it becomes known `post_type` ... and this happens quite late.
I will finish my tests and write more.

picasso:
My research has shown that `post_type` is only available after the` current_screen` action. In addition, there are situations when we need to disable script loading, but leave REST processing (for example, `rest_request_after_callbacks`). Therefore, it is possible to divide the "Gutenberg" support into two blocks:
- script enqueue and `rest_prepare`
- REST modifications (`rest_request_before_callbacks` and `rest_request_after_callbacks`)

But in the current situation, I managed with one filter. You can add 2 filters for the future, or you can make one filter and wait for the situation when someone needs 2 filters.

I suggest calling the filter `qtranslate_sl_block_editor` (Single Language edit mode)

herrvigg:
I've renamed to `qtranslate_admin_block_editor` to match the current need.
I've inverted boolean to signify "let QTX administrate the block editor". Therefore it becomes `true` by default and you should set it to `false` in your case.
Does it sound good?

picasso:
> Does it sound good?

Yes, it is quite suitable. I will change my logic from `true` to `false`

herrvigg (approved):

Issue #779, opened 2020-01-27
cookie vs. wp-cron.php


thabbs:
It looks like that qtranslate-xt tries to set the language cookie even when invoked by wp-cron.php. If there is no reason to do so it shouldn't try.

This warning is logged every time wp-cron is running:
```
Cannot modify header information - headers already sent in
[...]/wp-content/plugins/qtranslate-xt/qtranslate_core.php on line 463
```

herrvigg:
Which version are you using? The `set_cookie` call should be on line 473 not 463. The `qtranxf_can_redirect` is checking `wp_doing_cron` so i don't know, need more investigation. Check in debug mode, i also added some asserts that can help to understand.

herrvigg:
@thabbs I just updated Wiki [Troubbleshooting](https://github.com/qtranslate/qtranslate-xt/wiki/Troubleshooting). Maybe you could enable the WP_DEBUG_LOG to better see what's going on. I can't find the problem without any information from your side.

herrvigg:
@thabbs i can't reproduce this, so i close it for now. Feel free to re-open when you can provide more information.

herrvigg:
@thabbs This should be solved by #1114, just merged in master.

PR #1114, opened 2022-02-22
fix "Headers already sent" error in PHP

picasso:
`setcookie` is function modifying the HTTP header (need to send a `Set-Cookie:`)
sometimes __wp_cron.php__ seems to be the source of headers being sent prematurely
 so we avoid setting the cookie during `wp_doing_cron` if the headers have already been sent (it's painless for the plugin to **skip** this)

### DIFF
--- a/qtranslate_core.php
+++ b/qtranslate_core.php
@@ -447,6 +447,13 @@ function qtranxf_detect_language_front( &$url_info ) {
 function qtranxf_setcookie_language( $lang, $cookie_name, $cookie_path ) {
     global $q_config;
 
+    // fix "Headers already sent" error in PHP
+    // 'setcookie' is function modifying the HTTP header (need to send a 'Set-Cookie:')
+    // sometimes wp_cron.php SEEMS to be the source of headers being sent prematurely
+    // so we avoid setting the cookie during 'wp_doing_cron' if the headers have
+    // already been sent (it's painless for the plugin to skip this)
+    if(headers_sent() && wp_doing_cron()) return;
+
     // SameSite only available with options API from PHP 7.3.0
     if ( version_compare( PHP_VERSION, '7.3.0' ) >= 0 ) {
         setcookie( $cookie_name, $lang, [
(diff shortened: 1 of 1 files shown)


sviluppomania (approved):

qtranslate_core.php
```diff
@@ -447,6 +447,13 @@ function qtranxf_detect_language_front( &$url_info ) {
 function qtranxf_setcookie_language( $lang, $cookie_name, $cookie_path ) {
     global $q_config;
 
+    // fix "Headers already sent" error in PHP
+    // 'setcookie' is function modifying the HTTP header (need to send a 'Set-Cookie:')
+    // sometimes wp_cron.php SEEMS to be the source of headers being sent prematurely
+    // so we avoid setting the cookie during 'wp_doing_cron' if the headers have
+    // already been sent (it's painless for the plugin to skip this)
+    if(headers_sent() && wp_doing_cron()) return;
```
herrvigg, line 455: It makes sense to check for `headers_sent()`.
But I'm wondering, do we even have to check for `wp_doing_cron()`?
If the headers are sent we're already in troubles in any case. We would see the "Headers already sent" as a debug message, but it's pretty bad anyway. This should be detected differently, perhaps with an assert or some other functionality that we don't need to solve now.

herrvigg:
See also #779.

qtranslate_core.php
herrvigg: But we might also be conservative and simply fix this case as it was already reported in #779.
Then we can consider a "headers already sent" warning as a debug message for the general case. It doesn't seem to happen anymore, at least from QTX. If other plugins echo something before it's not depending on us.

picasso: Yes, I agree that `headers_sent()` is already enough.
But I inserted `wp_doing_cron()` for this reason:
I turned on debugging and controlled for a month. And all error messages were only when `wp_doing_cron()` was __true__.
If the headers have already been sent - it's an error. In another plugin or in the theme. And it means it should be seen and fixed. But `wp_cron.php` does not depend on us and won't be fixed. For me, it is logical to simply avoid this situation, especially since setting cookie during `wp_doing_cron()` there is no particular sense.

herrvigg: OK, let's keep it as it is. Merged.

PR #1115, opened 2022-03-02
QTS: Incorporated settings in qtranslate and cleanup

spleen1981:


modules/slugs/includes/qtranslate-slug-settings.php
```diff
@@ -486,109 +316,10 @@ function qts_validate_options( $input ) {
...
-                            )
-                        );
-
-                        $input[ $option['id'] ]       = trim( $input[ $option['id'] ] ); // trim whitespace
-                        $input[ $option['id'] ]       = force_balance_tags( $input[ $option['id'] ] ); // find incorrectly nested or missing closing tags and fix markup
-                        $input[ $option['id'] ]       = wp_kses( $input[ $option['id'] ], $allowed_html ); // need to add slashes still before sending to the database
-                        $valid_input[ $option['id'] ] = addslashes( $input[ $option['id'] ] );
-                        break;
-                }
-                break;
```
herrvigg, line 577: Nice cleanup! So we don't need all of these types anymore? They have not been used for a while it seems.

modules/slugs/includes/qtranslate-slug-settings.php
```diff
@@ -310,37 +186,44 @@ function qts_sanitize_bases( $base_slugs ) {
...
+            if ( $option['section']==$id ) {
+                qts_show_form_field( $option );
+            }
+        }
+    }
+    QTX_Admin_Settings::close_section( 'slugs' );
+    flush_rewrite_rules();
+}
+
+add_action( 'qtranslate_configuration', 'qts_show_settings_page');
```
herrvigg, line 211: Good trick!

modules/slugs/includes/qtranslate-slug-settings.php
```diff
@@ -310,37 +186,44 @@ function qts_sanitize_bases( $base_slugs ) {
...
+        <?php qts_section_fn($id);
+        // call the "add_settings_field" for each!
+        foreach ( $settings_output['qts_page_fields'] as $option ) {
+            if ( $option['section']==$id ) {
+                qts_show_form_field( $option );
+            }
+        }
+    }
+    QTX_Admin_Settings::close_section( 'slugs' );
+    flush_rewrite_rules();
```
herrvigg, line 208: I see this `flush_rewrite_rules` comes from the old form. Do we need this when showing the options? Shouldn't it rather be done on save?

herrvigg (commented):
Great job, nice cleanup!

modules/slugs/includes/qtranslate-slug-settings.php
spleen1981: Yes, not used and most probably not applicable to this module settings.
Other types are currently not used as well (e.g. `select`, `checkbox`, `multi-checkbox`), but maybe could be used in the future (validation logics to be reviewed in this case), so I left those there for now.
Additional cleanup could be needed when the module will be more mature and stable.

modules/slugs/includes/qtranslate-slug-settings.php
spleen1981: agree, ref 578e24f633225663823619506a2840d1ae042716

herrvigg: Great!

PR #1120, opened 2022-03-09
QTS: moved qts textdomain to qtranslate + general qtranslate pot/po/mo files re-alignment

spleen1981:


modules/slugs/includes/qtranslate-slug-settings.php
```diff
@@ -102,8 +102,8 @@ function qts_show_form_field( $args = array() ) {
         case 'select':
             echo "<select id='$id' class='select$field_class' name='" . QTS_OPTIONS_NAME . "[$id]'>";
             foreach ( $choices as $item ) {
-                $value = esc_attr( $item, 'qts' );
-                $item  = esc_html( $item, 'qts' );
```
herrvigg, line 106: It seems the old calls were wrong with the domain being sent as 2d argument, but these functions only take 1 arg.

modules/slugs/includes/qtranslate-slug-settings-options.php
```diff
@@ -6,8 +6,8 @@
  */
 function qts_options_page_sections() {
     $sections               = array();
-    $sections['post_types'] = __( 'Post types', 'qts' );
-    $sections['taxonomies'] = __( 'Taxonomies', 'qts' );
+    $sections['post_types'] = __( 'Post types', 'qtranslate' );
+    $sections['taxonomies'] = __( 'Taxonomies', 'qtranslate' );
```
herrvigg, line 10: We could have re-used 'Taxonomy' from WP default domain, but no big deal.

herrvigg:
Great!

PR #1121, opened 2022-03-09
QTS: rationalize get post_types/taxonomies functions and calls

spleen1981:


modules/slugs/includes/qtranslate-slug-settings-options.php
```diff
@@ -47,15 +47,27 @@ function qts_options_page_fields() {
         $options[] = qts_options_page_build_slug_fields( $taxonomy, "taxonomies", "taxonomy_" );
     }
 
-    return $options;
+    return array_filter($options);
```
herrvigg, line 50: What is that for?

modules/slugs/includes/qtranslate-slug-settings-options.php
```diff
@@ -47,15 +47,27 @@ function qts_options_page_fields() {
...
+            $slug = $object->name;
+        }else{
+            return array();
+        }
+    } else if (array_key_exists( 'slug', $object->rewrite ) ) {
+        $slug = ltrim( $object->rewrite['slug'], "/" );
+    } else {
+        $slug = $object->name;
+    }
+
```
herrvigg, line 65: Could we have objects that don't have `rewrite`?

modules/slugs/includes/qtranslate-slug-settings-options.php
```diff
@@ -47,15 +47,27 @@ function qts_options_page_fields() {
...
-    return $options;
+    return array_filter($options);
 }
 
 function qts_options_page_build_slug_fields( $object, $target_section, $id_prefix ) {
+    if ( ! is_array($object->rewrite) ) {
+        if ( $object->rewrite ){
+            $slug = $object->name;
+        }else{
+            return array();
```
herrvigg, line 58: Could we have objects that don't have the `rewrite` field here? Shouldn't they be eliminated earlier?

modules/slugs/includes/qtranslate-slug-settings-options.php
```diff
@@ -35,7 +35,7 @@ function get_multi_txt_choices() {
  * @return array
  */
 function qts_options_page_fields() {
-    $post_types = get_post_types( array( '_builtin' => false, 'public' => true ), 'objects' );
+    $post_types = get_post_types( array( 'public' => true ), 'objects' );
```
herrvigg, line 38: What is the effect of adding the builtin post types? Why were they filtered out before?

modules/slugs/includes/qtranslate-slug-settings-options.php
spleen1981: if `rewrite==false` for a post/taxonomy type, empty array is added in `qts_options_page_build_slug_fields` to `$options`, just cleaning up the array here.

modules/slugs/includes/qtranslate-slug-settings-options.php
spleen1981: yes e.g. for posts and pages `rewrite==false` meaning that permalink bases cannot be modified. `rewrite`property can be boolean or array. as long as `rewrite!=false` permalink base is ok to be customized

modules/slugs/includes/qtranslate-slug-settings-options.php
spleen1981: yes e.g. for posts and pages `rewrite==false` meaning that permalink bases cannot be modified. `rewrite`property can be boolean or array. as long as `rewrite!=false` permalink base is ok to be customized

modules/slugs/includes/qtranslate-slug-settings-options.php
spleen1981: last `else `is covering the case `rewrite `is an array but no `slug `key is there (still ok to be customized)

modules/slugs/includes/qtranslate-slug-settings-options.php
spleen1981: issue with some `_builtin` type (e.g. post and pages) is that slugs cannot be modified (hence customized slug will apparently just have no effect), but correct indication is not the `_builtin` type itself (which may include also types with slugs that could be modified), but the `rewrite=false` which is actually used to filter the post/taxonomy types in `qts_options_page_build_slug_fields` later.

herrvigg (approved):

PR #1122, opened 2022-03-10
QTS: install function rewritten without using $wpdb

spleen1981:


modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -108,17 +108,15 @@ public function get_meta_key( $force_lang = false ) {
...
     public function install() {
-        global $wpdb;
-
-        if ( function_exists( 'is_multisite' ) && is_multisite() ) {
-            if ( isset( $_GET['networkwide'] ) && ( $_GET['networkwide'] == 1 ) ) {
-                $old_blog = $wpdb->blogid;
-                $blogids  = $wpdb->get_col( $wpdb->prepare( "SELECT blog_id FROM $wpdb->blogs" ) );
-                foreach ( $blogids as $blog_id ) {
-                    switch_to_blog( $blog_id );
+        if ( is_multisite() ) {
```
herrvigg, line 114: Not needed, as it's called in `is_plugin_active_for_network`.

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -108,17 +108,15 @@ public function get_meta_key( $force_lang = false ) {
...
+            if ( is_plugin_active_for_network(plugin_basename(QTRANSLATE_FILE)) ) {
+                $old_blog = get_current_blog_id();
+                $blogs  = wp_get_sites();
```
herrvigg, line 117: Deprecated in WP 4.6.0 -> they say to use `get_sites`

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -108,17 +108,15 @@ public function get_meta_key( $force_lang = false ) {
...
+                foreach ( $blogs as $blog ) {
+                    switch_to_blog( $blog[blog_id]);
```
herrvigg, line 119: `blog_id` is not defined. I suppose you meant `$blog->id`. Not tested apparently.

herrvigg (changes requested):
Better with the WP API but there are a few things to fix.

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: agree

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: agree

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: my bad, could not test multisite with my current test environment... this just looked like straight forward.
`wp_get_sites()` should return arrays with key `blog_id`, while `get_sites()` is returning `WP_Site` arrays with property `blog_id`. Agree and changed to the latter

herrvigg (approved):

PR #1123, opened 2022-03-11
updated/completed italian plugin translation

spleen1981:
follows #1120


herrvigg (changes requested):
Can this be merged in #1120 directly? It's fine to add a new commit, I will squash anyway. If you don't manage I can do it, if this was the intention. I suppose it was just the PO/MO for the it_IT file.

spleen1981:
added to #1120, closing here

PR #1126, opened 2022-03-13
QTS: revision and cleanup of add/edit terms slugs input code

spleen1981:


modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -1476,9 +1476,10 @@ public function qts_modify_term_form( $term ) {
         echo "<script type=\"text/javascript\">\n// <![CDATA[\r\n";
         echo "
 		var slugforms = jQuery('#qts_term_slugs').html();
-		jQuery('#slug').parent().html(slugforms)\n;
-		console.log(slugforms);
-		
+                if ( jQuery('#slug').length ) {
+                    jQuery('#slug').parent().html(slugforms);
+                    jQuery('#qts_term_slugs').remove();
```
herrvigg, line 1481: Why is that `qts_term_slugs` even created in first hand? In which cases do we want this?

herrvigg: Why is that `slugforms` added in Javascript? Can't we simply keep `#qts_term_slugs` and update the fields on language change? That was a hack but it looks really weird.

spleen1981: `qts_term_slugs` is created as default input for taxonomies edit pages (e.g. for `edit-tags.php?taxonomy=category`), and original slug field is hidden.
For single term pages the `qts_term_slugs` does not fit well with the existing page layout, hence code is copied in place of the original slug input (if exists) placed in the existing table and `qts_term_slugs` is removed.
I guess the intent is just to reuse the same actions for both cases above.
As there is no dedicated hooks here (afaik) to make things cleaner, any possible approach would be hacky. Hence I retained the original philosophy...

herrvigg: OK I see. Perhaps there's a better way, as there are two different hooks (add/edit), but let's keep it like this for now.

However there's another part I don't understand, not coming from this patch but maybe you know:
https://github.com/qtranslate/qtranslate-xt/blob/dd241f42b6195049273c844d23f4d6d81899b695/modules/slugs/includes/class-qtranslate-slug.php#L1290-L1291

I can't find when this is triggered. When we edit a term it goes in the `else` part.

spleen1981: Ok, I think that that condition intention was the same currently accomplished with javascript.
This is not working (maybe with some old WP release it was), hence it is always evaluated `false `and maybe later the jquery patch was introduced and this has been left behind as junk code...
Initial intention was just to fix current situation, but actually this needs to be fully revised: I've added fe5961f2b504d70434e99b471e1f6ef350d5c273

- `qts_modify_term_form` removed
- callbacks for `_add_form_fields `and `_add_form_fields `are differentiated, no need for conditions
- only hack left is `hide_term_slug_box `which hides through jquery the original slugs fields, but there is no other way afaik
- removed apparently orphan function `qts_insert_term_input `

herrvigg: Great, this looks better! I have other comments but for that part we should be fine.

herrvigg:
I have some issues when saving a term, but I don't think it's coming just from this patch.

Warning: Undefined array key "es" in `.../includes/class-qtranslate-slug.php` on line 1346.
```
public function validate_term_slug( $slug, $term, $lang ) {
        global $q_config; //TODO: q_config  : term_name
        $term_key = qtranxf_split( $term->name );
        // after split we will get array (with language code as a key )
        $term_key = $term_key[ $this->default_language ];
        $name_in_lang = $q_config['term_name'][ $term_key ][ $lang ];
```

Here is category I'm testing
image

The `$term_key` becomes `Actualités` which is already quite weird, taking the name for the default language as a key.
The values are just set for FR, EN, DE in database, they are empty for IT and ES but here the form fields take the default language slug.

It's possible it doesn't handle well existing terms which are missing the slugs for every language.

herrvigg:
Also, another consideration: why don't we use the Language Switch Buttons (LSB) instead of duplicating all the fields for every language? With the LSB enabled we should use those and keep a single "slug" field to stick the general UI offered by QTX.

herrvigg:
Another minor bug, when we switch language with the LSB, the `slug` field (read-only) is not updated in the table.

image

I don't expect all of this to be solved in this Pull Request, but I leave it for info.

Let me know what you think about these 3 last remarks and see how we proceed. We can first merge this patch if you like.

herrvigg:
Another minor bug, when we switch language with the LSB, the `slug` field (read-only) is not updated in the table.

image

It should change to `news` for english and so forth.

herrvigg:
I don't expect all of this to be solved in this Pull Request, but I leave it for info.

Let me know what you think about these 3 last remarks and see how we proceed. We can first merge this patch if you like.

herrvigg:
Point 1 - can be worth looked at in this PR.
Point 2 - require certainly much more changes with JS code. Can be treated as a feature request.
Point 3 - could be related to point 2, as the read-only updates are also done in JS.

herrvigg:
Once this patch is merged, we should split this huge `class-translate-slug.php` file. It's way too big (1870 lines). The terms part could be taken out, among others.

spleen1981:
Point 1 should be covered by 1d2653ca831e10534d0bc07579c5d469c3004571, let me know if it works. I've removed unneeded complications which really did not seem to add value (besides the wrong implementation itself, with the missing check if key existed).
Point 2 & 3 need some reflection, as I think that slug metadata structure would need to be changed to use a single string same as translated content (e.g. [:en]slug1[:it]slug2[:]) rather than current mutiple strings, which is a big change in the module structure. Changing only javascript will only fix the LSB, but would not work in raw editor for example. However this makes sense and can be done, also for consistency with main plugin, but I would consider it as an enhancement and keep the matter on hold until the module is clean and stable. To formally fix the column bug for now, we may name the column "Slug (default language)".

herrvigg:
> Point 1 should be covered by https://github.com/qtranslate/qtranslate-xt/commit/1d2653ca831e10534d0bc07579c5d469c3004571, let me know if it works. I've removed unneeded complications which really did not seem to add value (besides the wrong implementation itself, with the missing check if key existed).

Yes it works, awesome!

herrvigg:
Points 2 and 3: it will require more work, indeed. We can create separate tickets to keep track of it.

PR #1135, opened 2022-03-27
generalized manual loaded modules mechanism

spleen1981:


herrvigg:
Good initiative! I had this on my todo-list.
As you probably noticed, I added you as member of the qTranslate organization :) So you could merge directly. But I'd like to review such Pull Requests, especially when impacting core functionalities. There's some WIP (Work In Progress) stuff and I have a few things in mind. I added a required approval in the config, but I don't know if this also applies to org members with write permission. Let's see.

herrvigg:
My idea was to have some menu/switches in the main list, so any module can be disabled. But it can certainly evolve from this change.

spleen1981:
> 

Ok just tried with an harmless commit (f363fd10dde21eb96cfd96fa10718a9b9755dae7) and it works as follows:
image

For next commits I will be waiting your review (my bad, I would have done like that in the first place...)

spleen1981:
> My idea was to have some menu/switches in the main list, so any module can be disabled. But it can certainly evolve from this change.

Makes sense, let me take a look.
Also some other structural modifications are needed in order to avoid hardcoding modules list/definitions in core files.
Definitions should be included in the single module folder itself following a scheme defined in core files along with the defaults, and list should be created based on mudules folder content. I will propose an implementation.
Other things will come up to mind while working I guess...

spleen1981:
> > My idea was to have some menu/switches in the main list, so any module can be disabled. But it can certainly evolve from this change.
> 
> Makes sense, let me take a look. 

maybe something like #1137

PR #1137, opened 2022-03-27
modules: manual activation check to main table

spleen1981:


admin/qtx_admin_settings.php
```diff
@@ -259,6 +259,7 @@ private function add_sections( $nonce_action ) {
         $admin_sections['import']      = __( 'Import', 'qtranslate' ) . '/' . __( 'Export', 'qtranslate' );
         $admin_sections['languages']   = __( 'Languages', 'qtranslate' );
 
+        //TODO: this actually assumes every manual activation module has settings, dedicated key to be added if that is not the case...
```
herrvigg, line 262: The activation setting has to be saved differently. I may revisit the `ma_module_enabled` config.

admin/qtx_admin_settings.php
```diff
@@ -743,7 +744,17 @@ class="qtranxs_explanation"><?php _e( 'The built-in integration modules are auto
...
+                                    <?php if ( isset( $q_config['ma_module_enabled'][ $module['id'] ] ) ) : ?>
+                                        <label for="ma_module_enabled_<?php echo $module['id']; ?>">
+                                            <input type="checkbox"
+                                                   name="ma_module_enabled[<?php echo $module['id']; ?>]"
+                                                   id="ma_module_enabled_<?php echo $module['id']; ?>"
+                                                   value="1"<?php checked( $q_config['ma_module_enabled'][ $module['id'] ] ) ?>/>
+                                            <?php echo $module['name']; ?>
+                                        </label>
+                                    <?php else: echo $module['name']; endif; ?>
+                                </td>
```
herrvigg, line 757: Good for now in but eventually I'd want a checkbox for all modules. But some have dependencies, I'll look how to evolve this.

modules/qtx_modules_handler.php
```diff
@@ -64,62 +64,60 @@ public static function update_manual_enabled_modules() {
...
+                'incompatible' => 'wp-seo-qtranslate-x/wordpress-seo-qtranslate-x.php',
             ),
             array(
                 'id'                => 'slugs',
                 'name'              => __( 'Slugs translation', 'qtranslate' ) . sprintf( ' (%s)', __( 'experimental' ) ),
                 'plugin'            => true,
                 'incompatible'      => 'qtranslate-slug/qtranslate-slug.php',
                 'manual_activation' => true,
-                'ma_checkbox_text'  => __( 'Enable slugs translation.', 'qtranslate' ),
-                'ma_checkbox_notes' => __( 'This activates the slug translation module. ', 'qtranslate' ) . __( 'Attention! This module is still experimental. It is subject to bugs and limitations.', 'qtranslate' )
```
herrvigg, line 122: Good to remove this.

herrvigg (approved):

admin/qtx_admin_settings.php
spleen1981: As long as it's only a matter of dependencies it should be enough to add `'manual_activation' => true` flag in configuration and it should work, the option will be just an additional condition to the dependencies (needs to be tested though).
But as those are integration plugin, maybe some of them are strictly needed for the relevant plugin to work correctly or at all with qtranslate, I don't know. If this is the case that should not be a setting exposed to the user.
If it's just a matter of layout, greyed-out checkboxes can be easily added following the internal plugin state.

herrvigg: There's some bug, I tried adding `manual_activation` to ACF just to try, and the checkbox doesn't appear.
If you know of a quick fix that would be good.

Looking forward, this looks too complicated to maintain, I'd like to simplify this without `ma_module_enabled` in `q_config`. This info should be better integrated with the `$module` object. I didn't expect so many changes regarding the modules before the next release.

spleen1981: pls check #1147

PR #1138, opened 2022-03-27
rationalize bool load options checks and extend to arrays

spleen1981:


herrvigg (commented):
I have to understand this better, need more time.

Not due to this patch but I'm not happy with the serialization of the options, it's such a big mess... really awful 😕 It's been like this for a very long time, but we should go back to something much simpler. We should use standard types and basic PHP serialization or JSON in the worst case. This would require a major rework though.

herrvigg:
As a side note, I updated the Wiki guidelines to add a link to the [best practices by Google](https://google.github.io/eng-practices/review/developer/) regarding the CL = ChangeList. It's become the facto standard in software engineering. It's really good advices.

In particular, I like the usage of **imperative tense** in the first line: https://google.github.io/eng-practices/review/developer/cl-descriptions.html#firstline e.g. "Rationalize ..." instead of "Rationalized" or "Rationalizing". It may sound as a nitpicking detail, but when you look later at the commit history it's much better and easier to understand the sequence of patches.

spleen1981:
> Not due to this patch but I'm not happy with the serialization of the options at all, it's such a big mess... really awful 😕 It's been like this for a very long time, but we should go back to something much simpler. We should use standard types and basic PHP serialization or JSON in the worst case. This would require a major rework though.

Agree, a cleanup and re-thinking would be great in the future.
However this PR is only reorganizing existing code and extending explicit checks on bools arrays instead of current handling, which directly loads arrays expected to be bool from db as strings and relies later on implicit casts (which currently happen to work, anyways).

qtranslate_core.php
```diff
@@ -621,14 +621,27 @@ function qtranxf_load_option_flag_location( $nm ) {
...
+            case 'y':
+            case 'yes':
+                return true;
+        }
+    } elseif (is_array( $var )){
+        $result=array();
+        foreach ($var as $key => $element){
+            $result[$key]=qtranxf_validateBool($element);
+        }
+        return $result;
```
herrvigg, line 642: I prefer not to mix basic/fundamental types with arrays here. So I'd abandon this part.

qtranslate_core.php
```diff
@@ -621,14 +621,27 @@ function qtranxf_load_option_flag_location( $nm ) {
...
+        switch ( $val ) {
+            case '0':
+            case 'n':
+            case 'no':
+                return false;
+            case '1':
+            case 'y':
+            case 'yes':
+                return true;
+        }
```
herrvigg, line 636: This looks better than the previous version, maybe we could keep just this part.
But It could be quite overkill, it's enough to "cast" strings like "1"/"0" with `(bool) val`. I don't know when other cases like `yes`/ `no` or even `y`/`n` are used. Need to check.

qtranslate_core.php
```diff
@@ -621,14 +621,27 @@ function qtranxf_load_option_flag_location( $nm ) {
     }
 }
 
-function qtranxf_validateBool( $var, $default_value ) {
```
herrvigg, line 624: In fact that function is unused in `master` (dead code).

herrvigg (commented):
Now I understand why you came up with this. It's mostly superseded by #1151 but some parts are/could still be valid.

spleen1981:
guess we can close this...

herrvigg:
Sorry for late answer, yes, I don't think we need it, but I find the existing code handling the options (from qTranslate / qTranslate-X) unnecessarily complicated. I'd like to refactor this, but it's been a lower priority.

PR #1140, opened 2022-03-27
modules: added flexible loading and removed hardcoded list

spleen1981:


herrvigg (commented):
Interesting idea, I understand the concept. It makes every folder self-defined.
However, I'm not convinced this is a good change:
- more code to maintain
- slower at compile time and runtime
- harder to maintain if we change the array structure, more files to change.

Such structure would be more relevant if we had to integrate new configs from external plugins, but it's not the case. The module structure is internal to QTX and I don't see the justification for making it more complex than it is.
We could have the config in a separate file, but I believe it's enough with a single file, kept very simple.

spleen1981:
Yes, this definitely has pros and cons...
The idea was to have modules development completely independent from core files (assumed the minimum required keys in the definitions array structure are not changing).
In this way those could be considered plugins and could be added without touching the core files upstream.
But I understand your point, in this case could be overkill...

herrvigg:
It would be relevant if we want to manage external modules.
The idea remains there, we can resurrect this code if want to expand this in the future.

PR #1142, opened 2022-03-28
QTS/WC: added slug fields to custom post type attributes pages

spleen1981:


herrvigg (commented):
Can we do it without adding new hooks (action/filter)? There's a cost in maintenance of adding hooks. This should be internal to QTX. No need to generalize for now I believe.

spleen1981:
> Can we do it without adding new hooks (action/filter)? There's a cost in maintenance of adding hooks. This should be internal to QTX. No need to generalize for now I believe.

done a1b99d3b51c4daeb6e5eb3e4e9507d0e73b025a3

modules/woo-commerce/qwc-admin.php
```diff
@@ -414,3 +414,9 @@ function qtranxf_wc_admin_filters() {
 }
 
 qtranxf_wc_admin_filters();
+
+
+function qtranxf_wc_admin_add_term_slugs($add_slugs_hooks){
+    return array_merge($add_slugs_hooks,
+            array('woocommerce_after_add_attribute_fields'));
+}
```
herrvigg, line 422: This is not called

modules/slugs/includes/qtranslate-slug-admin.php
```diff
@@ -30,6 +30,11 @@
 function qts_taxonomies_hooks() {
     global $qtranslate_slug;
 
+    $add_slugs_hooks=apply_filters( 'qts_add_tax_slugs_hook', array() );
+    foreach ( $add_slugs_hooks as $hook ){
+        add_action( $hook, 'qts_show_add_term_fields' );
+    }
+
```
herrvigg, line 37: The filter/action are still there.

herrvigg (changes requested):
The changes with the last commit seem incomplete. It doesn't look intentional.

modules/woo-commerce/qwc-admin.php
spleen1981: fixed, 47e425f76a21f14b8818d34425e19c6951964b91

modules/slugs/includes/qtranslate-slug-admin.php
spleen1981: fixed, 47e425f76a21f14b8818d34425e19c6951964b91

modules/slugs/includes/qtranslate-slug-admin.php
```diff
@@ -40,6 +40,10 @@ function qts_taxonomies_hooks() {
             add_filter( 'manage_' . $taxonomy->name . '_custom_column', 'qts_taxonomy_custom_column', 0, 3 );
         }
     }
+
+    if ( is_plugin_active( 'woocommerce/woocommerce.php' ) ) {
+        add_action( 'woocommerce_after_add_attribute_fields', 'qts_show_add_term_fields' );
+    }
```
herrvigg, line 46: This is the first time we have such combined integration. Shouldn't we check the WC module state instead?

spleen1981: That is exactly why I used the previous hook mechanism...

spleen1981: We may check the module status, but eventually the check will be the same.
At least in this way you are covered in the remote case you want to manually disable woocommerce module and still use the slugs.

herrvigg: Yes, the module state should be enough. It can be done without hooks, I want to avoid them as much as possible because it makes the maintenance more complicated.

spleen1981: done 6513fb44b3c3e45e5bcab871f12e719ca758c486

herrvigg (approved):
Looks good. I rewrote the module state check, better have it in the handler, the logic should stay there.

PR #1145, opened 2022-04-01
WC: removed unneeded action mistakenly used as a filter

spleen1981:


herrvigg:
Sounds good if you found this useful, but we should simply remove it in that case or mention it differently.
I want to avoid keeping commented code.
Btw how was this patch merged? Do you have access to the admin override?

spleen1981:
> Do you have access to the admin override?

Yes, ref. https://github.com/qtranslate/qtranslate-xt/pull/1135#issuecomment-1079921880 and https://github.com/qtranslate/qtranslate-xt/pull/1141#issue-1183180304

> Sounds good if you found this useless, but we should simply remove it in that case or mention it differently.
I want to avoid keeping commented code.

Agree, only reason I left it there was that I was missing the original intent (to find an alternate solution), but I'm ok with removing the comment.

PR #1146, opened 2022-04-03
Mods: display settings tabs from state

herrvigg:
The tab display is independent from its manual activation.
Any module could have extra settings (e.g. ACF).

Make the display dependent on the module state instead.
Add a new `has_settings` field in the module defintions.
Rename `load_modules_enabled` to `load_active_modules`.

### DIFF
--- a/modules/qtx_modules_handler.php
+++ b/modules/qtx_modules_handler.php
@@ -7,26 +7,43 @@ define( 'QTX_MODULE_STATUS_BLOCKED', 3 );
 
 class QTX_Modules_Handler {
     /**
-     * Loads modules previously enabled in the options after validation for plugin integration on admin-side.
+     * Get the modules previously activated in the options after validation for plugin integration on admin-side.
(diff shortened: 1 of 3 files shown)


modules/qtx_modules_handler.php
```diff
@@ -118,6 +135,7 @@ public static function get_modules_defs() {
                 'plugin'            => true,
                 'incompatible'      => 'qtranslate-slug/qtranslate-slug.php',
                 'manual_activation' => true,
+                'has_settings'      => true,
```
herrvigg, line 138: Note: I want to generalize the activation feature. Eventually `manual_activation` may disappear.
It's possible that `has_settings` will also evolve to store a callback for the settings to be called directly.

spleen1981: I was thinking the same thing... check #1147

spleen1981 (approved):

PR #1147, opened 2022-04-03
Mods: generalize manual activation

spleen1981:


modules/qtx_modules_handler.php
```diff
@@ -73,6 +75,7 @@ public static function update_manual_enabled_modules() {
         if ( $changed ) {
             update_option( 'qtranslate_modules', $options_modules );
             self::load_active_modules();
+            // TODO remove this action
```
spleen1981, line 78: We need to give opportunity to modules to perform activation/deactivation actions (e.g. flush rewrite rules for slugs module) on manual state change.
I don't see other clean ways than a `do_action` here...

herrvigg: I know it's needed, but using a hook for something purely internal is a *hack*. No need to fix this now but it can *always* be designed differently ;)
The hooks are the main reason why WordPress became so popular, and one of the reasons this framework is such a huge mess.

admin/qtx_admin_options_update.php
```diff
@@ -598,20 +598,28 @@ function qtranxf_update_setting( $var, $type = QTX_STRING, $def = null, $bool_el
...
+                if ( ! is_array( $val ) ) {
                     $val = sanitize_text_field( $val );
                     $val = preg_split( '/[\s,]+/', $val, -1, PREG_SPLIT_NO_EMPTY );
                 }
             } else {
-                $val = '';
+                $val = array();
             }
-            if ( empty( $val ) && ! is_null( $def ) ) {
+            if ( ! $bool_elements_array && empty( $val ) && ! is_null( $def ) ) {
```
herrvigg, line 608: This `$bool_elements_array` came from the patch I didn't review. It's making the function unnecessarily complicated and difficult to read. I will refactor in a new patch.

herrvigg (commented):
A few things fixed on the fly. Will merge but there will be other changes to come.

herrvigg (approved):
Good extension. There will be a follow-up.

admin/qtx_admin_options_update.php
```diff
@@ -598,20 +598,28 @@ function qtranxf_update_setting( $var, $type = QTX_STRING, $def = null, $bool_el
...
                     $val = preg_split( '/[\s,]+/', $def, -1, PREG_SPLIT_NO_EMPTY );
                 } else if ( is_array( $def ) ) {
                     $val = $def;
                 }
             }
+            // in case of bools array compare updating settings arrays with defaults to find keys set as false
+            if ( $bool_elements_array && ! is_null( $def ) && is_array( $def ) ) {
+                foreach ( $def as $key => $value ) {
+                    if ( ! array_key_exists( $key, $val ) ) {
+                        $val[ $key ] = false; //e.g. unchecked checkboxes input are not included in _$_POST
```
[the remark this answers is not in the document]
herrvigg: I had to revert as it didn't work as expected due to missing values for the checkboxes. 
The manipulation of this bool array is getting too complicated. More refactoring is needed to simplify this.

PR #1153, opened 2022-04-23
Refactor module classes

herrvigg:
The module objects are very difficult to manipulate with simple arrays.
Reorganize the code for a much better structure:
- `QTX_Admin_Module`: provides the basic static data (id, name, plugins, ...) and the built-in setup.
- `QTX_Admin_Module_Manager`: manages the module state updates and notices
- `QTX_Admin_Module_Settings`: holds the data relevant for the settings, mostly for display
- `QTX_Module_Loader`: reads the state and loads modules (used by admin and front core)
Move files to `modules` folder to keep the related code together.


modules/qtx_module_loader.php
```diff
@@ -0,0 +1,67 @@
...
+     * @see QTX_Admin_Modules::update_modules_state()
+     */
+    public static function get_active_modules() {
+        $options_modules = get_option( 'qtranslate_modules_state', array() );
+        if ( ! is_array( $options_modules ) ) {
+            return array();
+        }
+
+        $active_modules = array();
+        $modules        = QTX_Module_Setup::get_modules();
```
herrvigg, line 28: Mmm... I just realized we don't even need this in the module loader!
All the module setup and code complexity should only be for the admin.

modules/qtx_module_loader.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Loads modules previously activated in the options after validation for plugin integration on admin-side.
+     * Note these should be loaded before "qtranslate_init_language" is triggered.
+     *
+     * @see QTX_Admin_Modules::update_modules_state()
+     */
+    public static function load_active_modules() {
+        $modules = self::get_active_modules();
+        foreach ( $modules as $module ) {
+            include_once( QTRANSLATE_DIR . '/modules/' . $module->id . '/' . $module->id . '.php' );
```
herrvigg, line 64: We have everything in the options already, `id` and `state` are enough. The deps to the module setup can be broken completely. The goal is to make the loader as thin as possible for the front-end.

herrvigg (commented):
Refactoring helps to have a complete different view.

Issue #1156, opened 2022-04-28
Issue with slugs in Cyrillic


bagaweb:
I noticed an issue with latest experimental qTranslate-XT (with slugs module included):

when Russian language is enabled, Cyrillic slugs give 404 error.

If the slugs are in Latin characters (for example apartamenty instead of апартаменты), there are no issues.

It was working ok with qTranslate-XT 3.11.4 and qTranslate slug 1.1.8.

Anyone else noticed this bug?

spleen1981:
> It was working ok with qTranslate-XT 3.11.4 and qTranslate slug 1.1.8.

This should be broken since WP 4.8.3 due to a breaking change with `esc_sql()`

Fixed by #1157.

herrvigg:
@bagaweb Fix merged in master, please let us know if this also works on your side.

bagaweb:
Just trying the last update but unfortunately still 404 on the Russian version, actually now 404 errors are also in English and German version which are in Latin characters.

spleen1981:
I can't reproduce, it is working good with my test configuration...
Can you try to:
- double check slugs module activation status (maybe disable and enable again to be sure)
- update permalinks settings as explained [here](https://github.com/qtranslate/qtranslate-xt/tree/master/modules/slugs#i-get-a-404-error-what-can-i-do)
- update the test post before trying or test with a fresh one

herrvigg:
I'd like to release the new version but this part is blocking the release.
What are the steps and the config required to reproduce?

spleen1981:
@herrvigg to test this just set something like `пробный` as an alternate language slug for a post/page, with slugs module activated, and try to reach the url.

herrvigg:
Good, it seems to works fine.

@bagaweb If it fails for other languages the issues are most likely not due to Slugs and not even QTX. Perhaps some problem with URL rewrites in your web server (Apache, nginx, ...).
Try to disable all and see if WP works with raw values. Then reactivate QTX and see where it fails. With QTX be sure you use the correct URL values before the custom slug values, according to the URL mode.

herrvigg:
I can also confirm that fix is doing the job.
- If I go back before fix (`git checkout f3889ab02d495969378fbbb837343680e3a82ff4`) I can reproduce the error. Page not found.
- With the fix the problem disappears.

bagaweb:
Hello @herrvigg & @spleen1981  I'm back to testing and still getting the same error with latest master.

[...]

PR #1157, opened 2022-05-01
QTS: fix to handle esc_sql() 4.8.3 breaking change

spleen1981:
Since 4.8.3, ‘%’ characters will be replaced with a placeholder string with esc_sql().
This fix replaces back the placeholder to ‘%’ after esc_sql() call to retain the original behaviour.

### DIFF
--- a/modules/slugs/includes/class-qtranslate-slug.php
+++ b/modules/slugs/includes/class-qtranslate-slug.php
@@ -957,7 +957,7 @@ class QtranslateSlug {
         $page_path     = str_replace( '%20', ' ', $page_path );
         $parts         = explode( '/', trim( $page_path, '/' ) );
         $parts         = array_map( 'esc_sql', $parts );
-        $parts         = array_map(array($wpdb,'remove_placeholder_escape'), $parts);
+        $parts         = array_map( array( $wpdb, 'remove_placeholder_escape' ), $parts );
         $parts         = array_map( 'sanitize_title_for_query', $parts );
         $in_string     = "'" . implode( "','", $parts ) . "'";
         $meta_key      = QTS_META_PREFIX . $this->get_temp_lang();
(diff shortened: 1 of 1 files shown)


herrvigg:
Great if this fixes #1156! I trust you on this, I haven't a test case ready. Maybe @bagaweb could also test this before merge (easy to copy-paste the fix locally)?

Could you update the TODO in the readme if all is solved?

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -957,6 +957,7 @@ private function get_page_id_by_path( $page_path, $output = OBJECT, $post_type =
         $page_path     = str_replace( '%20', ' ', $page_path );
         $parts         = explode( '/', trim( $page_path, '/' ) );
         $parts         = array_map( 'esc_sql', $parts );
+        $parts         = array_map( array( $wpdb, 'remove_placeholder_escape' ), $parts );
         $parts         = array_map( 'sanitize_title_for_query', $parts );
```
herrvigg, line 961: The 3x `array_map` may be regrouped in one function (even anonymous) for faster execution in a sngle `array_map`.

spleen1981:
> Could you update the TODO in the readme if all is solved?

This fixes only #1156 which is relevant to posts/pages slugs.
TODO list points are relevant to taxonomies, that's a separate issue, I will take a look.

@bagaweb you can apply this [patch](https://github.com/qtranslate/qtranslate-xt/commit/908d05487bc557efef6959e7c1721b16391d4d14.patch) for a quick test.

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: 5ca8b04995b336a292be9aa2a8b50d4f8e440dc5

herrvigg (approved):

PR #1158, opened 2022-05-01
Harmonize module options

herrvigg:
The module custom options are difficult to track due to legacy names.
Make all the name consistent by renaming the old options:
- `acf_qtranslate` -> `qtranslate_module_acf`
- `qts_options` -> `qtranslate_module_slugs`
- `qtranslate_modules` -> `qtranslate_modules_state`
  (already pre-renamed in #1150)

Generalize migration and cleanup on plugin activation:
- legacy options belonging to external plugins are preserved and imported
- internal options are renamed from legacy + cleanup.

Create new constants to better keep track the names (e.g. ACF forms) and deletion.
New custom options for modules are updated with `autoload=false`.
Add temporary import for `master`, to be removed before release.

### DIFF
--- a/admin/qtx_activation_hook.php
+++ b/admin/qtx_activation_hook.php
@@ -753,7 +753,6 @@ function qtranxf_clear_debug_log() {
     }
 }
 
-
 function qtranxf_activation_hook() {
     qtranxf_clear_debug_log();
     if ( version_compare( PHP_VERSION, '5.4' ) < 0 ) {
@@ -784,6 +783,11 @@ function qtranxf_activation_hook() {
     if ( is_plugin_active( 'qtranslate-x/qtranslate.php' ) ) {
         qtranxf_admin_notice_deactivate_plugin( 'qTranslate-X', 'qtranslate-x/qtranslate.php' );
     }
+
+    // Migrate legacy options, temporary transitions during evolutions.
+    qtranxf_migrate_legacy_option( 'qtranslate_modules', QTX_OPTIONS_MODULES_STATE );
+    qtranxf_migrate_legacy_option( 'acf_qtranslate', QTX_OPTIONS_MODULE_ACF, false );
+    qtranxf_migrate_legacy_option( 'qts_options', QTX_OPTIONS_MODULE_SLUGS, false );
 
     $ts                     = time();
     $next_thanks            = get_option( 'qtranslate_next_thanks' );
--- a/admin/qtx_activation_hook.php
+++ b/admin/qtx_activation_hook.php
@@ -784,10 +784,10 @@ function qtranxf_activation_hook() {
         qtranxf_admin_notice_deactivate_plugin( 'qTranslate-X', 'qtranslate-x/qtranslate.php' );
     }
 
-    // Migrate legacy options, temporary transitions during evolutions.
-    qtranxf_migrate_legacy_option( 'qtranslate_modules', QTX_OPTIONS_MODULES_STATE );
-    qtranxf_migrate_legacy_option( 'acf_qtranslate', QTX_OPTIONS_MODULE_ACF, false );
-    qtranxf_migrate_legacy_option( 'qts_options', QTX_OPTIONS_MODULE_SLUGS, false );
+    // Migrate (rename/import) legacy options, temporary transitions during evolutions.
+    qtranxf_rename_legacy_option( 'qtranslate_modules', QTX_OPTIONS_MODULES_STATE );
+    qtranxf_import_legacy_option( 'acf_qtranslate', QTX_OPTIONS_MODULE_ACF, false );
+    qtranxf_import_legacy_option( 'qts_options', QTX_OPTIONS_MODULE_SLUGS, false );
 
     $ts                     = time();
     $next_thanks            = get_option( 'qtranslate_next_thanks' );
--- a/admin/qtx_admin_options.php
+++ b/admin/qtx_admin_options.php
@@ -91,3 +91,28 @@ function qtranxf_admin_load_config() {
 
     qtranxf_add_conf_filters();
 }
+
+/**
+ * Migrate, rename and clean up a legacy option.
+ *
+ * Recopy the legacy option if the new doesn't already exist.
+ * Delete the legacy option if belonging to `qtranslate`, preserve external plugin options.
+ *
+ * @param string $old_name
+ * @param string $new_name
+ * @param bool|string $autoload as in update_option
+ *
+ * @return void
+ */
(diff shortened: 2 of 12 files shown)


admin/qtx_activation_hook.php
```diff
@@ -785,6 +784,11 @@ function qtranxf_activation_hook() {
         qtranxf_admin_notice_deactivate_plugin( 'qTranslate-X', 'qtranslate-x/qtranslate.php' );
     }
 
+    // Migrate legacy options, temporary transitions during evolutions.
+    qtranxf_migrate_legacy_option( 'qtranslate_modules', QTX_OPTIONS_MODULES_STATE );
+    qtranxf_migrate_legacy_option( 'acf_qtranslate', QTX_OPTIONS_MODULE_ACF, false );
+    qtranxf_migrate_legacy_option( 'qts_options', QTX_OPTIONS_MODULE_SLUGS, false );
```
herrvigg, line 790: Maybe I should separate migration from import.

herrvigg: Don't merge yet, I will make it clearer.

herrvigg: Done.

admin/qtx_admin_options.php
```diff
@@ -91,3 +91,39 @@ function qtranxf_admin_load_config() {
...
+/**
+ * Import legacy option by recopying its value if the new option doesn't already exist.
+ *
+ * @param string $old_name
+ * @param string $new_name
+ * @param bool|string $autoload as in update_option
+ *
+ * @return void
+ */
+function qtranxf_import_legacy_option( $old_name, $new_name, $autoload = null ) {
```
spleen1981, line 104: Should this function return `update_option` result (and `true` for other cases)?

admin/qtx_admin_options.php
```diff
@@ -91,3 +91,39 @@ function qtranxf_admin_load_config() {
...
+ * @param string $new_name
+ * @param bool|string $autoload as in update_option
+ *
+ * @return void
+ */
+function qtranxf_rename_legacy_option( $old_name, $new_name, $autoload = null ) {
+    assert( strpos( $new_name, 'qtranslate_' ) === 0 );
+    assert( strpos( $old_name, 'qtranslate_' ) === 0 );
+    qtranxf_import_legacy_option( $old_name, $new_name, $autoload );
+    delete_option( $old_name );
```
spleen1981, line 128: Better test  `qtranxf_import_legacy_option` before doing this.

admin/qtx_admin_options.php
```diff
@@ -91,3 +91,39 @@ function qtranxf_admin_load_config() {
...
+ *
+ * @param string $old_name
+ * @param string $new_name
+ * @param bool|string $autoload as in update_option
+ *
+ * @return void
+ */
+function qtranxf_import_legacy_option( $old_name, $new_name, $autoload = null ) {
+    assert( strpos( $new_name, 'qtranslate_' ) === 0 );
+    if ( ! get_option( $new_name ) ) {
```
spleen1981, line 106: Not sure new named option should always have the priority. 
Maybe we could test also module activation state (test above `AND` module is active) though maybe involves a too remote scenario to complicate the function...

qtranslate_options.php
```diff
@@ -55,10 +55,22 @@
 // Language code format: ISO 639-1 (2 alpha), 639-2 or 639-3 (3 alpha)
 define( 'QTX_LANG_CODE_FORMAT', '[a-z]{2,3}' );
 
+/**
+ * Option names.
+ */
+const QTX_OPTIONS_MODULES_STATE = 'qtranslate_modules_state';
+const QTX_OPTIONS_MODULE_ACF    = 'qtranslate_module_acf';
+const QTX_OPTIONS_MODULE_SLUGS  = 'qtranslate_module_slugs';
```
spleen1981, line 63: These definitions would look better in modules source files.
It would be better to segregate code which depends on module list from core files.

admin/qtx_admin_options_update.php
```diff
@@ -360,7 +360,9 @@ function qtranxf_reset_config() {
     // internal private options not loaded by default
     delete_option( 'qtranslate_next_update_mo' );
     delete_option( 'qtranslate_next_thanks' );
-    delete_option( 'qtranslate_modules_state' );
+    delete_option( QTX_OPTIONS_MODULES_STATE );
+    delete_option( QTX_OPTIONS_MODULE_ACF );
+    delete_option( QTX_OPTIONS_MODULE_SLUGS );
```
spleen1981, line 365: These would look better somewhere in modules source files to make things cleaner, maybe a new function to be called from here.
It would be better to segregate code which depends on module list from core files.

admin/qtx_admin_options.php
herrvigg: With a rename the old option should be deleted in any case or if it it appears again.
If the new option exist, it should not be overwritten. But the old option should removed even if nothing was updated.

admin/qtx_admin_options.php
herrvigg: I haven't found a case where I'd need it.

admin/qtx_admin_options.php
herrvigg: The idea of the import is a one-shot action at activation, the goal is that QT-XT cares of its own options (modules).

The normal scenario is that new options don't exist yet. If we can import a value such as ACF or slugs legacy settings (coming from original plugins) the new option is created from these values to ensure the transition from the old plugins. But from that point we want to remain with the QT-XT setup and ignore the legacy stuff once we have all set in QT-XT.

The most tricky case is when the QT-XT options are reset, we don't really want to import again but in that case we expect the admin to save a new set of options right after the reset. If this is not done, a new import can happen again.

This patch is only valid for the modules as it's a new feature in QT-XT but this will be expanded to all other options later once we cut completely the links with QT-X (not XT).

qtranslate_options.php
herrvigg: The idea is that all options definitions should be kept consistent in a unique place (the options names are used both for front and admin). Currently it's hard to to understand where a general option is set or read in the code because the values are spread everywhere and most of the settings are generated from intermediate strings (with  `qtranslate_` prefix). The modules should not be fundamentally different regarding the options interface. The other options should also converge to something more comprehensive.

admin/qtx_admin_options_update.php
herrvigg: I also thought this, but this belongs more to general admin functionalities for the whole QT-XT plugin. For the same reason as for the option names, better keep those with the rest and not split them to avoid the risks of missing updates/deletions.

spleen1981 (approved):

Issue #1159, opened 2022-05-01
QTS: permalink bases with UTF8 characters not working


spleen1981:
Confirmed issue in slugs module from legacy qtranslate-slugs plugin README.md, reported here for better tracking and visibility

> In slug options you can change the bases for taxonomies and custom post types.
So, for example, you can change /category/ for /category/ for english and /categoria/ for spanish version. But these won't work:
>- slug with UTF8 characters in taxonomies bases: example: /類別/... instead of /category/... UTF8 in taxonomies works just fine: /category_zh/魚/
>- slug with UTF8 characters in custom post type bases : example: /圖書/... instead of /books/... UTF8 in custom post slugs works just fine: /tushu/彩繪中國經典名著/
>- translating custom post types archives with custom base name /tushu/ isn't working. Using UTF8 in the the default slug works as expected : /中國/

herrvigg:
Merged in master.

PR #1160, opened 2022-05-02
QTS: fix query var handling, utf8 bases support

spleen1981:
* Fix logic breaking query variables getter in filter_request. Former logic overwrote $query arg with $wp->matched_query. Current one merges the two (preserving possible query vars) only if needed (404 is detected) and removes only the error key.
* Fix language switch links in case of 404 to point to home page instead of current not existing resource.
* Reorder tests and incorporate in single if block
* Add support for utf8 bases (fallback to $wp->request).


spleen1981:
Fixes #1159

herrvigg (approved):
I can't test all of this but the changes look fine for slugs, great with new fixes.

PR #1161, opened 2022-05-03
WC: fix attribute edit page hidden fields

spleen1981:
Display the set of QTS slugs when editing (or adding) a WC product attribute.
Hide the regular slug.

### DIFF
--- a/modules/woo-commerce/qwc-admin.php
+++ b/modules/woo-commerce/qwc-admin.php
@@ -44,8 +44,6 @@ function qtranxf_wc_add_filters_admin() {
 
     add_filter( 'woocommerce_attribute_taxonomies' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
     add_filter( 'woocommerce_variation_option_name' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
-    //TODO: check upstream in QTX core if a cleaner approach can be adopted
-    add_action( 'woocommerce_after_edit_attribute_fields', function(){echo "<script>jQuery(document).ready(function(){jQuery(\"td\").show()})</script>";});
 }
 
 qtranxf_wc_add_filters_admin();
--- a/modules/woo-commerce/qwc-admin.php
+++ b/modules/woo-commerce/qwc-admin.php
@@ -41,9 +41,11 @@ function qtranxf_wc_add_filters_admin() {
     foreach ( $email_common as $name => $priority ) {
         add_filter( $name, 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage', $priority );
     }
-    
+
     add_filter( 'woocommerce_attribute_taxonomies' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
     add_filter( 'woocommerce_variation_option_name' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
+    //TODO: check upstream in QTX core if a cleaner approach can be adopted
+    add_action( 'woocommerce_after_edit_attribute_fields', function(){echo "<script>jQuery(document).ready(function(){jQuery(\"td\").show()})</script>";});
 }
 
 qtranxf_wc_add_filters_admin();
--- a/modules/slugs/includes/qtranslate-slug-admin.php
+++ b/modules/slugs/includes/qtranslate-slug-admin.php
@@ -44,6 +44,7 @@ function qts_taxonomies_hooks() {
 
     if ( QTX_Module_Loader::is_module_active( 'woo-commerce' ) ) {
         add_action( 'woocommerce_after_add_attribute_fields', 'qts_show_add_term_fields' );
+        add_action( 'woocommerce_after_edit_attribute_fields', 'qts_show_edit_term_fields' );
     }
 }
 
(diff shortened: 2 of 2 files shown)


modules/woo-commerce/qwc-admin.php
```diff
@@ -41,9 +41,11 @@ function qtranxf_wc_add_filters_admin() {
     foreach ( $email_common as $name => $priority ) {
         add_filter( $name, 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage', $priority );
     }
-    
+
     add_filter( 'woocommerce_attribute_taxonomies' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
     add_filter( 'woocommerce_variation_option_name' , 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
+    //TODO: check upstream in QTX core if a cleaner approach can be adopted
+    add_action( 'woocommerce_after_edit_attribute_fields', function(){echo "<script>jQuery(document).ready(function(){jQuery(\"td\").show()})</script>";});
```
herrvigg, line 48: Mmm yes we'd need something better. But how is this supposed to work? How is that code executed in the browser? Why just selecting any `td` element? Hard to understand.

spleen1981: This is injected in one specific page only (woocommerce attribure edit screen) which has just a few labels and tds containing one input each. For some reason tds are hidden (hence page is unusable) due to some qtx js from upstream. Once the page is loaded the injected code shows up the tds again.
Considering that it's a very specific issue and modifying general upstream js may cause issues somewhere else, this workaround should be enough for now. However finding the actual root cause in the qtx js part it's not exactly trivial, it would take some time while the feature above is broken,

herrvigg: > For some reason tds are hidden (hence page is unusable) due to some qtx js from upstream

Let's look at this better. It seems we can do a much cleaner fix

spleen1981: Ok, in the meanwhile I have filled in #1162 for tracking.
This PR can be merged as a temporary workaround to guarantee usability or we can just leave the thing on hold while a better solution comes out, your choice.

herrvigg: See
https://github.com/qtranslate/qtranslate-xt/issues/1162#issuecomment-1120239358

So this hook is simply undoing what QTS is doing earlier.
We can delay the release 3.12.0 until this is clarified, would be nice to have since Slugs seems quite well integrated with WC otherwise.

herrvigg: If the solution to handle the attributes is to hard to implement with slugs active, let's keep the fields hidden for now.
But the WC script to undo what QTS is doing (specifically for WC) is not what we'd want for sure.

spleen1981: Ouch, I completely forgot we moved WC stuff into slugs module, good catch! Those js workarounds are evil...
Reverted previous commit and modified existing QTS workaround with [d3e7688](https://github.com/qtranslate/qtranslate-xt/pull/1161/commits/d3e7688769fd00ec839fd26e2f457f1cc4f4f122)

herrvigg (commented):
OK this looks better, but when we edit a product attribute, we just get one slug field.
Shouldn't we have all the set of QTS slugs for each language?

herrvigg:
Shouldn't that `qts_show_edit_term_fields` also apply to WC product attributes?

herrvigg:
This seems to work:
```
function qts_taxonomies_hooks() {
    ...
    if ( QTX_Module_Loader::is_module_active( 'woo-commerce' ) ) {
        add_action( 'woocommerce_after_add_attribute_fields', 'qts_show_add_term_fields' );
        add_action( 'woocommerce_after_edit_attribute_fields', 'qts_show_edit_term_fields' );  <--- new
    }`
```

herrvigg:
Oh but we have the LSB switch buttons, that's a bit confusing having the list + switches. But I don't know how this is supposed to work, I haven't really tested it yet.

spleen1981:
> OK this looks better, but when we edit a product attribute, we just get one slug field. Shouldn't we have all the set of QTS slugs for each language?

Support for multi slugs seems was missing for this page.
added here https://github.com/qtranslate/qtranslate-xt/pull/1161/commits/fd168ee156770038ef577318418f542a657d4a45

> Oh but we have the LSB switch buttons, that's a bit confusing having the list + switches. But I don't know how this is supposed to work, I haven't really tested it yet.

LSB is only working for Name field, not for slugs here or in any other page. This needs to be implemented, but it's a big change in slugs module, it will need time to figure out and test. It's a 'to do' for next releases.

herrvigg (approved):

herrvigg:
I've rewritten a bit the functions to clarify the signatures. I'm afraid this doesn't work as expected for the WP attributes.
Problem on edit:
- the usual [edit_form_fields action ](https://developer.wordpress.org/reference/hooks/taxonomy_edit_form_fields/) sends the WP_Term
- the woocommerce_after_edit_attribute_fields action doesn't. If you search in the code, you can find this `do_action( 'woocommerce_after_edit_attribute_fields' );`

We need the term ID to retrieve the values from the meta, but I don't know if this also applies to WP product attributes (where are they stored?). Related to this, the `qts_save_term` is supposed to work with product attributes, might be something there as well (add/edit)

herrvigg:
Imo if we can't find a fix, better hide the language slugs until a solution is found.

herrvigg:
Also note that [add_form_fields action](https://developer.wordpress.org/reference/hooks/taxonomy_add_form_fields/) sends a `taxonomy` (string), not a WP_Term object. I have clarified this in this commit https://github.com/qtranslate/qtranslate-xt/commit/5b46dfffc452acd3fe4f19b524294d9622ff67a9.
With WC the action `do_action( 'woocommerce_after_add_attribute_fields' );` doesn't send anything, but we don't really need. What matters is `qts_save_term` but it's not triggered, neither for add or for edit WC attribute.

spleen1981:
Yes it seems this is not applicable for custom taxonomies base slugs.
However this is already handled in slugs module through modules settings, I think it is enough to redirect there to set up the multilingual slugs, same as it is done for in-built base slugs taxonomies (e.g. posts `category`).
Ref #1164

PR #1163, opened 2022-05-08
Improve slugs layout with flags

herrvigg:
Redesign the lists of slugs with flags and less text, for more compact layout and better styles with CSS:
- post metabox
- slugs settings
- terms (add/edit)

image


spleen1981 (approved):
Looks great!

PR #1167, opened 2022-05-14
Prevent slugs metadata removal on legacy plugin uninstallation

spleen1981:


herrvigg:
This is a temporary hack but it doesn't solve the problem well. We should own all these data in QTX with proper meta prefix.
I have a WIP solution in this branch, still some a bit to do but you can check it.
https://github.com/qtranslate/qtranslate-xt/tree/slugs-import

herrvigg:
Superseded by #1171.

PR #1168, opened 2022-05-15
QTS: fix conflict pages/posts in filter_request

spleen1981:
As the information that query is relevant to a page instead of a post is only included in `$wp->matched_query`, merge with `$query` argument must always occur.
If page is identified, `name` key is removed as it is used to retrieve posts instead.
Posts identification test is moved at the end of the block to avoid other conflicts with more specific tests (e.g. custom post types).

### DIFF
--- a/modules/slugs/includes/class-qtranslate-slug.php
+++ b/modules/slugs/includes/class-qtranslate-slug.php
@@ -353,6 +353,11 @@ class QtranslateSlug {
         global $q_config;
         global $wp;
 
+        /* As $wp->matched_query filters the query including slugs mudule custom rewrite rules, it provides all necessary data to reach the intended resource.
+         * However discarding completely $query arg and using only $wp->matched_query would result in losing possible custom query vars.
+         * Hence $query and $wp->matched_query are merged and unneeded/conflictual keys (e.g. 'error') from $query are removed.
+         */
+
         if ( isset( $wp->matched_query ) ) {
             if ( isset( $query['error'] ) ){
                 unset( $query['error'] );
@@ -372,6 +377,9 @@ class QtranslateSlug {
 
         // -> page
         elseif ( isset( $query['pagename'] ) || isset( $query['page_id'] ) ):
+            /* 'name' key from passed $query var is removed if 'pagename' key is present in $wp->matched_query.
+             * Both keys causes conflict between posts and pages as 'name' is used for posts, resulting in 404 error.
+             */
             if ( isset( $query['name'] ) ) {
                 unset($query['name']);
             }
@@ -421,6 +429,8 @@ class QtranslateSlug {
 
         else:
 
+            // If none of the conditions above are matched, specific tests to identify custom post types and taxonomies are performed here.
+
             // -> custom post type
             foreach ( $this->get_public_post_types() as $post_type ) {
                 if ( array_key_exists( $post_type->name, $query ) && ! in_array( $post_type->name, array(
@@ -464,6 +474,12 @@ class QtranslateSlug {
                     $function             = 'get_term_link';
                 }
             endforeach;
+
+            /* As 'name' key is present also at least both for pages and custom post types, this condition alone cannot be used to identify uniquely the posts.
+             * For pages and custom post types specific tests can be and are performed earlier but no additional specific condition seems to be applicable for posts.
+             * Given the above, the following test needs to be placed after the custom post types one, and is it positive if 'name' key is there and no other match is found earlier.
+             * If a specific condition was found to uniquely identify posts, this block should be placed in the main if block.
+             */
 
              // -> post
             if ( ! $function && ( isset($query['name']) || isset($query['p'] ) ) ) {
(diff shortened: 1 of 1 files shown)


modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -471,6 +464,19 @@ function filter_request( $query ) {
                     $function             = 'get_term_link';
                 }
             endforeach;
```
herrvigg, line 466: Why are the taxonomies in the main "if" inside the block for the custom type? This looks wrong. It wasn't so in the origin plugin.

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -471,6 +464,19 @@ function filter_request( $query ) {
...
+                $post = isset($query['p']) ? get_post($query['p']) : $this->get_page_by_path($query['name'], OBJECT, 'post');
+                if (!$post) {
+                    return $query;
+                }
+                $query['name'] = $post->post_name;
+                $id = $post->ID;
+                $cache_array = array($post);
+                update_post_caches($cache_array);
+                $function = 'get_permalink';
+            }
```
herrvigg, line 479: Why is that if-block nested inside the last else? The checks are not clear, it's getting too confusing. Is that for custom post type or more? Nested "if" should be really justified. It should be rewritten with very clear if sections. I preferred the former version with the "post" in the previous place.

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -370,6 +372,9 @@ function filter_request( $query ) {
 
         // -> page
         elseif ( isset( $query['pagename'] ) || isset( $query['page_id'] ) ):
+            if ( isset( $query['name'] ) ) {
+                unset($query['name']);
+            }
```
herrvigg, line 377: Why is the `name` removed?

modules/slugs/includes/class-qtranslate-slug.php
```diff
@@ -353,8 +353,10 @@ function filter_request( $query ) {
         global $q_config;
         global $wp;
 
-        if ( isset( $query['error'] ) && isset( $wp->matched_query ) ) {
-            unset( $query['error'] );
+        if ( isset( $wp->matched_query ) ) {
+            if ( isset( $query['error'] ) ){
+                unset( $query['error'] );
+            }
```
herrvigg, line 359: I don't understand what we are doing here with `matched_query`. The `error` was not checked in the original plugin. Comments are welcome because it's not obvious.

herrvigg (changes requested):
Good if this fixes the issue with the page but the code is difficult to read. Many questions. Should be rewritten in a clearer way.

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: If none of the conditions above are matched, custom post types first and taxonomies later are tried.
Original plugin tried those for every query, which was not needed.

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: Problem is that `name` key is passed in `$query` arg for pages, posts and custom post types at least.
Need to perform different checks on other keys or `matched_query` to identify which is the correct callback.
As `post` seems to involve the more generic test, it is left as last fallback if no other function has been identified from the previous tests.
Moving it in the previous position would always give priority on the custom post types, which will result in 404.
If we revert everything to the original code, `$query` is completely discarded braking other possible custom query vars (only `matched_query `was used).

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: both posts and pages have as a minimum only `page` and `name` keys. No way to distinguish between them looking only at `$query` arg only.
Must look into `matched_query` for `pagename` key to identify a page, but if `$query` is returned both with `name` and `pagename` keys, still posts have the priority and you get a 404.

modules/slugs/includes/class-qtranslate-slug.php
spleen1981: Ref above replies: `error` key was not checked originally because `$query` arg was discarded and replaced completely with `matched_query`, losing all possible additional custom query vars.
In previous commit I tried to limit the use of `matched_query` to the cases where nothing could be found with `$query` only (i.e. `'error' => 404`), but it seems in certain cases (e.g. pages) `matched_query` data is strictly needed.
Hence `matched_query` is now always merged with `$query` removing from the latter specific keys (e.g. `error`);

spleen1981:
> code is difficult to read
> Should be rewritten in a clearer way.

Agreed, but code reflects the strategy to identify the kind of item processed by current query (at least my understanding), which is actully not very linear, plus the need to preserve possible custom additional query vars.
Coulnd't came up with anything clearer, but I'm obviously open to changes as long as it works.

Added some clarifications here 3e954f6f42cec947ad8dd31f84e5038777dd50e3

modules/slugs/includes/class-qtranslate-slug.php
herrvigg: OK

herrvigg (approved):

PR #1171, opened 2022-05-22
Import QTS slugs options and meta data into QTX

herrvigg:
When slugs is uninstalled all QTS data are erased and will be lost for QTX. The options and meta data from the legacy plugin should be properly imported in QTX to separate entries.

Add new features in the import settings to migrated QTS data:
- QTS `postmeta` / `termmeta` slugs are migrated renaming `_qts_slug_` prefix to `qtranslate_slug_` prefix.
- QTS `qts_options` slugs are migrated to `qtranslate_module_slugs` options, after removing the `_qts_` prefix for each type entry.

ATTENTION: existing slugs options and meta before migration are erased! A confirmation is required. Without this, a dry-run mode allows to see the effects before change.

Update prefix constants accordingly for QTX:
- rename `QTS_META_PREFIX` to `QTX_SLUGS_META_PREFIX`
- delete `QTS_PREFIX` made obsolete.

Add a new admin notice to suggest import when old QTS options and meta are found. The import can be done manually by the admin at any time.
Remove most of the HTML `id` attributes from the Slugs settings form, not needed (`name` attributes matter).


herrvigg:
@spleen1981 please check this carefully on your side and let me know how it goes before merging it.

**Attention**: the current options that may have been changed during the last commits in master (stored in `qtranslate_module_slugs`) will be erased during import. The legacy QTS options are unchanged in any case.

This solves several issues of migration on activation which are hard to maintain and may create problems in the future. The main idea here is that manual import from QTS is **required** before being able to use slugs in QTX.

modules/slugs/includes/qtranslate-slug-settings.php
```diff
@@ -171,7 +171,7 @@ function qts_show_form_field( $args = array() ) {
                     $checked = 'checked="checked"';
                 }
 
-                echo "<label for='$id|$item_value'><input class='radio$field_class' type='radio' id='$id|$item_value' name='" . QTX_OPTIONS_MODULE_SLUGS . "[$id]' value='$item_value' $checked /> <strong>$item_key</strong>";
+                echo "<label for='qtx_slugs_$id'><input class='radio$field_class' type='radio' id='qtx_slugs_$id' name='" . QTX_OPTIONS_MODULE_SLUGS . "[$id]' value='$item_value' $checked /> <strong>$item_key</strong>";
```
herrvigg, line 174: I removed most of the HTML `id` attributes from the Slugs settings form because they created confusions.
They are not needed, unlike the `name` attributes that matter for the form POST.

admin/qtx_activation_hook.php
```diff
@@ -932,6 +931,32 @@ function qtranxf_admin_notices_gutenberg() {
 
 add_action( 'admin_notices', 'qtranxf_admin_notices_gutenberg' );
 
+function qtranxf_admin_notices_slugs_import() {
+    if ( qtranxf_check_admin_notice( 'slugs-import' ) ) {
+        return;
+    }
```
herrvigg, line 937: We should only check if the `slugs` module is active.

modules/slugs/includes/qtranslate-slug-import.php
```diff
@@ -0,0 +1,149 @@
...
+    if ( $new_options ) {
+        if ( $db_commit ) {
+            delete_option( QTX_OPTIONS_MODULE_SLUGS );
+        }
+        $msg[] = sprintf( __( "Deleted %s types from options.", 'qtranslate' ), count( $new_options ) );
+    }
+
+    $options = get_option( QTX_SLUGS_LEGACY_QTS_OPTIONS_NAME );
+    if ( $options ) {
+        $new_options = [];
```
spleen1981, line 115: Not clear why we want to delete existing new named options.
Shouldn't be better to merge existing new options array and renamed old options array to preserve possible new entries added between QTX update and migration?

modules/slugs/includes/qtranslate-slug-import.php
```diff
@@ -0,0 +1,149 @@
...
+     * Generic function that imports old meta into new ones. All existing new meta are erased first.
+     *
+     * @param string $table name of the meta table (postmeta, termmeta)
+     * @param string $colid column name giving the meta id (post_id, term_id)
+     * @param string[] $msg array of messages, updated
+     *
+     * @return void
+     */
+    $import_meta = function ( $table, $colid, &$msg ) use ( $wpdb, $old_prefix, $new_prefix ) {
+        $results = $wpdb->query( "DELETE FROM $table WHERE meta_key like '$new_prefix%'" );
```
spleen1981, line 71: Not clear why we want to delete existing new named metadata.
Ref similar comment done for existing new named options.

admin/qtx_import_export.php
```diff
@@ -229,6 +229,21 @@ function qtranxf_admin_section_import_export( $request_uri ) {
             'text' => sprintf( __( 'Use plugin %s to import data.', 'qtranslate' ), '<a href="https://wordpress.org/plugins/w2q-wpml-to-qtranslate/" target="_blank">W2Q: WPML to qTranslate</a>' )
         ) ) ?>
         <?php do_action( 'qtranslate_add_row_migrate' ) ?>
+        <tr id="qtranslate-import-slugs">
+            <th scope="row"><?php _e( 'Import from slugs', 'qtranslate' ) ?></th>
+            <td>
+                <label for="qtranslate_import_slugs"><input type="checkbox" name="qtranslate_import_slugs"
+                                                            id="qtranslate_import_slugs"
+                                                            value="1"/> <?php _e( 'Import options, post and term meta from legacy slugs (QTS). Attention, existing slugs will be deleted.', 'qtranslate' ); ?>
```
spleen1981, line 237: Maybe we could hide the confirmation checkbox unless the first is checked.
We may add inline to the first input element something like `onclick="let x=jQuery('#qtranslate_import_slugs_confirm');if (jQuery(this).prop('checked')===true){x.parent().show();}else{x.prop('checked', false);x.parent().hide();}"`.

modules/slugs/includes/qtranslate-slug-import.php
```diff
@@ -0,0 +1,149 @@
...
+ * Import slugs meta by duplicating the legacy QTS postmeta and termmeta for QTX.
+ *
+ * @param bool $db_commit true to commit changes, false for dry-run mode.
+ *
+ * @return string messages giving details.
+ */
+function qtranxf_slugs_import_qts_meta( $db_commit ) {
+    global $wpdb;
+
+    $new_prefix = QTX_SLUGS_META_PREFIX;
```
spleen1981, line 58: If options import is requested while slugs module is inactive `QTX_SLUGS_META_PREFIX` is undefined, resulting in fatal error .
Checkbox can be shown conditionally, `QTX_SLUGS_META_PREFIX` defined outside the module  or function can return with notice if test is negative.

modules/slugs/includes/qtranslate-slug-import.php
```diff
@@ -0,0 +1,149 @@
...
+                              SELECT $colid, REPLACE(meta_key, '$old_prefix', '$new_prefix'), meta_value
+                              FROM  $table
+                              WHERE meta_key like '$old_prefix%'" );
+        $msg[]   = sprintf( __( "Imported %s rows into $table (%s->%s).", 'qtranslate' ), $results ?: '0', $old_prefix, $new_prefix );
+    };
+
+    $msg = [];
+    $wpdb->query( "START TRANSACTION" );
+    $import_meta( $wpdb->postmeta, 'post_id', $msg );
+    $import_meta( $wpdb->termmeta, 'term_id', $msg );
```
spleen1981, line 85: Just realized that old metadata is not removed from $wpdb->termmeta on [legacy plugin uninstallation](https://github.com/not-only-code/qtranslate-slug/blob/386b90723d79792ffe4b88138908a065e82b01f2/qtranslate-slug.php#L182).
We may take care here and remove the old ones after successful import or add a checkbox to remove the orphaned data.

spleen1981: Fix needed also for current qtx uninstallation action, added in e91398359930a03d84fbf7e14fbad1a003c4d08b

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: QTX should not alter any of the slugs created by QTS. Those belong to QTS, it's another plugin. We should see them as read-only. What happens to those are not our concern.
We need to delete existing meta with new key in case they have already been imported once. If we repeat the import multiple times, duplicates would be created.

herrvigg (commented):
Good remarks

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Another option could also be to add "Delete existing `slugs` data before import". Then do a merge in any case.

admin/qtx_import_export.php
herrvigg: Yes, or maybe better, make it disabled (grayed) and enable it when the first one is checked.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: I mean add an additional option checkbox "Delete existing slugs data before import". Then it's all up to the admin to decide what he wants. In any case we don't touch the original QTS data so it's pretty safe.

spleen1981: > Perhaps the merge could still be done, if we consider it like a PHP array_merge. The new data will only replace the old data for new keys. The old keys are kept only if they don't exist in the new array.

Yes that is what I meant. I would avoid additional checkboxes, better doing it automatically.

modules/slugs/includes/qtranslate-slug-import.php
spleen1981: > However it could be possible to do a merge if we do it in two steps: (1) overwrite (update) the meta_value for existing meta_key (new prefix), and (2) insert missing keys (new prefix).

Yes, sounds good IMO

modules/slugs/includes/qtranslate-slug-import.php
spleen1981: covered by 5ed9b2e7088b0ca2f408a5e49bf366bfd7d36baf

admin/qtx_import_export.php
spleen1981: agreed, implemented in 5ed9b2e7088b0ca2f408a5e49bf366bfd7d36baf

modules/slugs/includes/qtranslate-slug-import.php
spleen1981: We should at least mention in import/module notes.
I submitted to original plugin repo a [PR](https://github.com/not-only-code/qtranslate-slug/pull/156) as it seems something has been merged recently.
In case it is merged we can mention that better if plugin unistallation is done from that commit on, otherwise those entries will need to be deleted manually from db.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: The new import now selects only the QTS slugs that have not been imported yet. Existing slugs (after rename) are ignored.

Added also a new option to delete before import. This gives max flexibility in case the admin wants to erase all and restart the import from scratch.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: The new import now selects only the QTS slugs that have not been imported yet. Existing slugs (after rename) are ignored.

Added also a new option to delete before import. This gives max flexibility in case the admin wants to erase all and restart the import from scratch.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Finally opted for (2), only new keys are imported. 
(1) can become problematic because it's hard to understand the purpose of the import. Instead, added a new option to delete before import. If checked, the import will clone all the QTS data. It is only meaningful in case the import is done multiple times.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: I see. The problem is that the current users may not have this change in their QTS plugin. Perhaps we could make a special case here and add a new option to delete explicitly the QTS meta after import (also taking the commit/rollback flag).

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Done. Finally opted for (2), the new import now selects only the QTS slugs that have not been imported yet. Existing slugs (after rename) are ignored. It was a bit tricky as the DB primary key is not the meta key.

Overwrite (1) can become problematic because it's hard to understand the purpose of the import in that case. So there's never any overwrite (SQL REPLACE never applies because the IDs are always different). 

Instead, added a new checkbox to _delete before import_. This gives max flexibility in case the admin wants to erase all and restart the import from scratch. It is only meaningful in case the import is done multiple times. It is to ask the import to clone all the QTS data.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Done. It's a bit more tricky than it seems but it should work.
Existing meta_key are preserved unless deletion is asked explicitly. If not, a conservative merge is done:
- for new types, all slugs are imported
- for existing types, slugs are merged, existing slugs are preserved (no overwrite).

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: I don't know what to do here. Adding a QTS cleanup is possible but it would make it too complicated maybe.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Need to re-open this. Problem: the QTX options are saved before the QTS import is run. So new keys are created and stored in new options before the import happens. As the import is conservative preserves existing values it doesn't do anything. Need more checks on values and overwrite empty slugs strings.

herrvigg:
@spleen1981 I'm starting to think this is getting too complicated for what we need... and the problem of QTS cleanup is not solved yet. What about doing a import-migration instead?
1. delete all existing QTX config about slugs
2. import options and meta from QTS by **renaming** the keys.

No selective import, it's all or nothing. It looks like my first solution but the QTS legacy is not kept anymore, it's a full transfer.

Pros:
- simpler: less code, clearer consequence of the action 
- solves the problem of QTS cleanup at the same time as all is moved. No leftovers in QTS.

Cons:
- the migration is **irreversible** from the admin panel. If old keys wanted to be restored it's possible but it has to be done by hand in SQL. But normally there's literally no reason to continue using QTS after import. The labels will made clearer. Admin should still do a DB backup before.
- current changes done on master branch would be **lost**, but that affects very few people (developers like you who are testing slugs). Imo not worth handling this in the regular source base. The changes should be limited as the QTS values still exist at this point.

modules/slugs/includes/qtranslate-slug-import.php
spleen1981: We could add a `register_uninstall_hook` for the legacy plugin with the missing action only `$wpdb->query("DELETE from $wpdb->termmeta WHERE meta_key IN ($meta_keys)");` in

spleen1981:
@herrvigg this solution is much better IMO, cons are not an issue as the legacy plugin is dead and is an external module anyways. I liked this approach more from the beginning, the simpler the better.

modules/slugs/includes/qtranslate-slug-import.php
herrvigg: Yes, but it should become unnecessary now with the new import-migration.
It can be added in a separate PR if needed, but I close this topic here to make the review clearer.

herrvigg:
OK - finally! The new version is much simpler, it becomes a migration as described above:
1. delete all existing QTX config about slugs
2. import options and meta from QTS by renaming the keys.

There's a step 0, I added a temporary hack for master dev! There was a nasty side effect of saving new options with new keys leading to empty strings. The hack restores the previous QTS options as before so the migration falls back to the general case again (I didn't think about meta data at that time).  It's really ugly but this will go away before release. Normal QTS users don't need this.

The patch should be ready for review. For security, save your `qtranslate_module_slugs` before confirming migration. The meta data are never lost, easy to recover if you want to repeat the migration. In that case just rename the meta keys the other way around:
```
UPDATE wp_postmeta
SET meta_key = REPLACE(meta_key, 'qtranslate_', '_qts_') 
WHERE meta_key like 'qtranslate_%';

UPDATE wp_termmeta
SET meta_key = REPLACE(meta_key, 'qtranslate_', '_qts_') 
WHERE meta_key like 'qtranslate_%';
```
Let me know how it goes.

modules/slugs/admin/slugs-migrate-qts.php
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Generic function that migrates QTS meta to QTX meta.
+     *
+     * @param string $table name of the meta table (postmeta, termmeta)
+     * @param string[] $msg array of messages, updated
+     *
+     * @return void
+     */
+    $migrate_meta = function ( $table, &$msg ) use ( $wpdb, $old_prefix, $new_prefix ) {
+        $results = $wpdb->query( "DELETE FROM $table WHERE meta_key like '$new_prefix%'" );
```
spleen1981, line 71: I think we should count here the number of `QTX_SLUGS_LEGACY_QTS_META_PREFIX` entries and return if it's 0.
It is pointless to remove current data if nothing is there to be imported, it is likely a mistaken operation, and out of function intent anyways.

modules/slugs/admin/slugs-migrate-qts.php
```diff
@@ -0,0 +1,153 @@
...
+function qtranxf_slugs_migrate_qts_options( $db_commit ) {
+    $msg = [];
+
+    $old_options = get_option( QTX_OPTIONS_MODULE_SLUGS );
+    if ( $old_options ) {
+        if ( $db_commit ) {
+            delete_option( QTX_OPTIONS_MODULE_SLUGS );
+        }
+        $msg[] = sprintf( __( "Deleted %s types from options.", 'qtranslate' ), count( $old_options ) );
+    }
```
spleen1981, line 108: I think this part should be moved inside the `if ( $qts_options ) {`, no reason to remove anything if no old option is there to be migrated.

modules/slugs/admin/slugs-migrate-qts.php
```diff
@@ -0,0 +1,153 @@
...
+    if ( $db_commit ) {
+        update_option( QTX_OPTIONS_MODULE_SLUGS, $new_options, false );
+        delete_option( QTX_SLUGS_LEGACY_QTS_OPTIONS_NAME );
+
+        global $qtranslate_slugs;
+        if ( $qtranslate_slugs->options_buffer != $new_options ) {
+            $qtranslate_slugs->options_buffer = $new_options;
+            flush_rewrite_rules();
+        }
```
spleen1981, line 127: This should be moved in `if ( $qts_options ) {` block as well.

modules/slugs/admin/slugs-migrate-qts.php
spleen1981, line 108: I think this part should be moved inside the `if ( $qts_options ) {`, no reason to remove anything if no old option is there to be migrated.

modules/slugs/admin/slugs-migrate-qts.php
```diff
@@ -0,0 +1,153 @@
...
+    }
```
spleen1981, line 128: This should be moved in `if ( $qts_options ) {` block as well.

modules/slugs/admin/slugs-migrate-qts.php
herrvigg: It's tricky to count before the DELETE is done. The rename must be done after the DELETE, otherwise there might be some conflicts with the key names. Some keys may already exist, some not. The admin should use the dry-run mode to see the effects before commit.

modules/slugs/admin/slugs-migrate-qts.php
herrvigg: Done. Early exit if no options to migrate.

modules/slugs/admin/slugs-migrate-qts.php
herrvigg: Done. Early exit if no options to migrate

modules/slugs/admin/slugs-migrate-qts.php
herrvigg: But perhaps it doesn't matter to know if all will be migrated. If there's no QTS meta at all, we can assume we don't want to delete anything, it's fine. I will push a change.

herrvigg: Add pre-check QTS for conditional delete and migrate by table. It's safer now.

herrvigg: In fact I made a confusion with the previous conservative import. In the new case it's much simpler, it becomes all or nothing. So it makes perfectly sense to count QTS data before deletion in QTX. Good point!

modules/slugs/admin/slugs-migrate-qts.php
```diff
@@ -0,0 +1,160 @@
...
+
+    $msg = [];
+    $wpdb->query( "START TRANSACTION" );
+    $migrate_meta( $wpdb->postmeta, $msg );
+    $migrate_meta( $wpdb->termmeta, $msg );
+    if ( $db_commit ) {
+        $wpdb->query( "COMMIT" );
+    } else {
+        $wpdb->query( "ROLLBACK" );
+    }
```
spleen1981, line 91: This will work for InnoDB tables only. If table is myISAM changes will be committed even in dry-run mode.
We should assume both scenario are possible.
To cover both we could just count `$new_prefix` and `$old_prefix` to give stats if dry run mode is requested, or fire directly the queries above otherwise (without any explicit START TRANSACTION/COMMIT/ROLLBACK).

herrvigg: Good point, I didn't think about that 😕 
Yes it can rewritten differently with pre-counts, now that the migration is simpler. Also, the DELETE can be more narrowed to the post/meta ID that will be renamed.

herrvigg: Rewrote the dry-run mode without SQL transaction. 
Add additional filter to narrow the DELETE query by groups of post_id/meta_id to be migrated. Other groups of slugs not concerned by the migration are left unchanged. It should change nothing from a pure migration from scratch but it makes it a bit safer if the admin runs it again later (though it should not).

herrvigg:
@spleen1981 let me know how the new version looks and I can merge it with an updated commit message.
After that I think we are finally ready for release 3.12.0.

spleen1981 (approved):
Looks good!

PR #1176, opened 2022-05-28
Remove internal QTS filter hooks

herrvigg:
The QTS filters come from the legacy plugin:
- filter: `qts_validate_post_slug`
- filter: `qts_validate_term_slug`
- filter: `qts_permastruct`
- filter: `qts_url_args`

For easier maintenance and readability, remove those filters and replace them with direct calls instead.
New `qtranslate` filters will be added on request if the need is justified.

Rename the two `validate` slugs functions to `sanitize` (post/term).
Minor cleanup to improve phpdoc and code readability.

### DIFF
--- a/modules/slugs/admin/slugs-admin.php
+++ b/modules/slugs/admin/slugs-admin.php
@@ -5,10 +5,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 include_once( dirname( __FILE__ ) . '/slugs-settings.php' );
 
 // add filters
-add_filter( 'qts_validate_post_slug', 'qts_validate_post_slug', 0, 3 );
-add_filter( 'qts_validate_post_slug', 'qts_unique_post_slug', 1, 3 );
-add_filter( 'qts_validate_term_slug', 'qts_validate_term_slug', 0, 3 );
-add_filter( 'qts_validate_term_slug', 'qts_unique_term_slug', 1, 3 );
 add_filter( 'wp_get_object_terms', 'qts_get_object_terms', 0, 4 );
 add_filter( 'get_terms', 'qts_get_terms', 0, 3 );
 // admin actions
@@ -143,15 +139,15 @@ function qts_draw_meta_box( $post ) {
 }
 
 /**
- * Sanitize title as slug, if empty slug.
+ * Sanitize a post slug for a given language.
  *
- * @param $post (object) the post object
- * @param $slug (string) the slug name
- * @param $lang (string) the language
+ * @param string $slug slug name
+ * @param WP_Post $post the post object
+ * @param string $lang the language
  *
- * @return string the slug validated
+ * @return string sanitized slug
  */
-function qts_validate_post_slug( $slug, $post, $lang ) {
+function qts_sanitize_post_slug( $slug, $post, $lang ) {
     $post_title = trim( qtranxf_use( $lang, $post->post_title ) );
     $post_name  = get_post_meta( $post->ID, QTS_META_PREFIX . $lang, true );
     if ( ! $post_name ) {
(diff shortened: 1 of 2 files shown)


modules/slugs/admin/slugs-admin.php
```diff
@@ -143,38 +139,36 @@ function qts_draw_meta_box( $post ) {
...
     if ( ! $post_name ) {
         $post_name = $post->post_name;
     }
 
     //TODO: if has a slug, test and use it
-    //TODO: and then replace the default slug with the dafault language slug
-    $name = ( $post_title === '' ) ? $post_name : $post_title;
-
+    //TODO: and then replace the default slug with the default language slug
+    $name = empty( $post_title ) ? $post_name : $post_title;
```
spleen1981, line 159: This is not covering values of `$post_title` that would be evaluated as `false`, like `'0'`. Better the previous test.

modules/slugs/admin/slugs-admin.php
```diff
@@ -143,38 +139,36 @@ function qts_draw_meta_box( $post ) {
...
     $slug = trim( $slug );
-
-    $slug = ( $slug === '' ) ? sanitize_title( $name ) : sanitize_title( $slug );
+    $slug = empty( $slug ) ? sanitize_title( $name ) : sanitize_title( $slug );
```
spleen1981, line 161: This is not covering values of `$slug` that would be evaluated as `false`, like `'0'`. Better the previous test.

modules/slugs/admin/slugs-admin.php
```diff
@@ -269,53 +263,55 @@ function qts_save_postdata( $post_id, $post = null ) {
...
- * @return string the slug validated
+ * @return string sanitized slug
  */
-function qts_validate_term_slug( $slug, $term, $lang ) {
+function qts_sanitize_term_slug( $slug, $term, $lang ) {
     global $q_config;
 
     $term_name = trim( qtranxf_use( $lang, $term->name, false, true ) );
-    if ( $term_name === '' ) {
+    if ( empty( $term_name ) ) {
```
spleen1981, line 298: This is not covering values of `$term_name` that would be evaluated as `false`, like `'0'`. Better the previous test.

modules/slugs/admin/slugs-admin.php
```diff
@@ -269,53 +263,55 @@ function qts_save_postdata( $post_id, $post = null ) {
...
         $term_name = trim( qtranxf_use( $q_config['default_language'], $term->name ) );
     }
     $slug = trim( $slug );
-    $slug = $slug === '' ? sanitize_title( $term_name ) : sanitize_title( $slug );
+    $slug = empty( $slug ) ? sanitize_title( $term_name ) : sanitize_title( $slug );
```
spleen1981, line 302: This is not covering values of `$slug` that would be evaluated as `false`, like `'0'`. Better the previous test.

modules/slugs/admin/slugs-admin.php
herrvigg: I don't `qtranxf_use` can ever return `false`. But this was not the main purpose of this patch, so I reverted this for clarity.

modules/slugs/admin/slugs-admin.php
herrvigg: Reverted (though `false` should never happen).

modules/slugs/admin/slugs-admin.php
herrvigg: Reverted (though `false` should never happen).

modules/slugs/admin/slugs-admin.php
herrvigg: Reverted.

herrvigg:
Reverted the empty string checks. It should never be `false` with `qtranxf` but that's for the post title / term name. For the input slugs there might be special cases with `false` that I don't see well yet. This can probably be rewritten but it's a different topic, it would be for another patch.

spleen1981 (approved):

Issue #1116, opened 2022-03-06
Coupon codes are not properly translatable


terzag:
Using qTranslate-XT 3.11.4 with the WooCommerce plugin enabled. WordPress 5.9.1 / WooCommerce 6.2.1.

When editing a coupon code, I can translate the post title (which shouldn't be the case as it's used for the coupon code) but not the description, as updating it in one language and switching to another one keeps the same content. I.e. after adding it to language A, switching to language B has the same one; after changing it in language B and switching back to language A, it has the same description as language B.

If coupons codes are to be translatable, it should be the other way around: no translation for title but translatable description.

I've seen that there were other similar bug reports from years ago, from which I understood that you filtered the coupons admin to not be translatable, so I'm wondering if that choice has been reverted or if there have been changes in WooCommerce that breaks your filtering?

herrvigg:
Related to #883 and #61. The coupons should not be translatable from what I understood. But it's not entirely clear. I'll let you check these discussions, if you have a clearer idea let us know.

spleen1981:
For the time being, post type `shop_coupon` should be manually disabled from qtranslate advanced settings.
Ref. also https://github.com/qtranslate/qtranslate-xt/issues/608#issuecomment-420820843
Automatism to "black list" specific types where translation would break the functionality, based on relevant module plugin info, may need to be added to cover properly.

PR #1192, opened 2022-06-16
Don't show slug metabox on woo shop_orders and shop_coupon pages

erichk4:
The slug metaboxes might not be necessary for the woocommerce orders and coupon pages...

### DIFF
--- a/modules/slugs/admin/slugs-admin.php
+++ b/modules/slugs/admin/slugs-admin.php
@@ -9,10 +9,7 @@ add_filter( 'wp_get_object_terms', 'qtranxf_slugs_get_object_terms', 0, 4 );
 add_filter( 'get_terms', 'qtranxf_slugs_get_terms', 0, 3 );
 
 // admin actions
-
-// add slug metabox
 add_action( 'add_meta_boxes', 'qtranxf_slugs_add_slug_meta_box' );
-
 add_action( 'save_post', 'qtranxf_slugs_save_postdata', 605, 2 );
 add_action( 'edit_attachment', 'qtranxf_slugs_save_postdata' );
 add_action( 'created_term', 'qtranxf_slugs_save_term', 605, 3 );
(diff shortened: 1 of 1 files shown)


modules/slugs/admin/slugs-admin.php
```diff
@@ -7,8 +7,16 @@
 // add filters
 add_filter( 'wp_get_object_terms', 'qtranxf_slugs_get_object_terms', 0, 4 );
 add_filter( 'get_terms', 'qtranxf_slugs_get_terms', 0, 3 );
+
 // admin actions
-add_action( 'add_meta_boxes', 'qtranxf_slugs_add_slug_meta_box' );
+
+// add slug metabox
+$curr_post_type = $_GET[ 'post_type' ] ? $_GET[ 'post_type' ] : get_post_type( intval( $_GET[ 'post' ] ) );
+if ( $curr_post_type !== 'shop_order' && $curr_post_type !== 'shop_coupon' )
```
herrvigg, line 15: Is that enough to read the `$_GET` values?

herrvigg (commented):
Looks good to me.
@spleen1981 please could you review? As you have integrated WC with slugs you could have better feedbacks.

I had to rewrite without the null coalescing operator as this is PHP7.0 and QTX still allows PHP 5.4+. The requirements will be raised soon, but until then the retro-compatibility is needed.

spleen1981:
Both `post `and `post_type` are not guaranteed to exist in $_GET, so they would need to be tested to avoid warnings.
Also `post_type` seems not to be the right key to test in this case.
Anyways `global $post_type` seems a cleaner test here.
I think also that is better to move the logic inside the callback, there is no real advantage doing otherwise and readability is worst IMO.
Also I think is better to have an array of "not applicable types" there, as I suspect others may come out.
The above implemented here https://github.com/qtranslate/qtranslate-xt/pull/1192/commits/edcb9ac0fe38c69550e1e2d39c2edd0208c12609

@herrvigg as a side note these post types should not be translatable at all (ref. also #1116), we should check upstream and grey-up unchecked the relevant post types checkboxes in qtranslate advanced settings.

modules/slugs/admin/slugs-admin.php
spleen1981: no, ref next comment

herrvigg (approved):

Issue #1179, opened 2022-06-02
Woocommerce webhooks


tonysilva16:
Hello,

With a variable product, if i update the variation price and click on save changes, woocommerce will send two webhooks, one of them have the raw data (parent product update), and the other webhook (variation update) all the fields are translated.

Example:
https://i.imgur.com/IBz5MiE.png

I already tried with the latest published version, and also with the current plugin files on the master branch.

Thanks.

spleen1981:
Not clear to me.
Can you explain in detail:

- what is the expected result
- what is the actual result instead

Also please copy/paste complete code/outputs here.

spleen1981:
@tonysilva16 can you test this [branch](https://github.com/qtranslate/qtranslate-xt/tree/wc_webhook_fix)

tonysilva16:
Hello @spleen1981 ,

For what i could test, one of the problems remains. When the product is a variation type, the returning data has translated attributes.
I installed xdebug and the data used in the webhook is retrieved in **get_product_data** function inside **WC_REST_Products_V2_Controller** class.

The function in the screenshoot bellow is called to get the attributes field.

![image](https://user-images.githubusercontent.com/65446091/174839819-c4982cad-b18c-4409-ab92-233aadbb3b8a.png)

The result of the get_attribute_taxonomy_name function is this one, that contains the translated attribute

![image](https://user-images.githubusercontent.com/65446091/174840955-f8384c28-ac89-4d6a-b5a2-61caadc5a80b.png)

If the product type is not a variation, the attribute name is correct.

![image](https://user-images.githubusercontent.com/65446091/174842050-e508332e-700f-43a3-9763-4e26a9e4bfdb.png)

spleen1981:
ok, it seems undes some circumnstances (e.g. variations updates) the webhook is generated through AJAX instead of cron, and in that case `qwc-admin.php` is loaded instead of `qwc-front.php` (would not work anyways as needed filter is added only if `wp_doing_cron()` is `true`).
Added a fix in the same branch above, let me know if it's working now.

tonysilva16:
Hello,

Its seems that is fixed, at least for what i could tried it was working allright. Tomorrow i am gonna test it better and i will give you feedback.
Thanks.

tonysilva16:
Hello,

I am getting an error on a coupon update. I am not sure, but the webhook must be trigger using the admin page and the webhook function is using a function only available when doing frontend

```
[...]

PR #1194, opened 2022-06-22
fix unintended translations in webhooks

spleen1981:
fix for  #1179

### DIFF
--- a/modules/woo-commerce/qwc-admin.php
+++ b/modules/woo-commerce/qwc-admin.php
@@ -4,8 +4,16 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 function qtranxf_wc_add_filters_admin() {
-    // priority 20 is used because in case other plugins add some untranslated content on normal priority
+ // priority 20 is used because in case other plugins add some untranslated content on normal priority
     // it will still hopefully then get translated.
+    $admin_hooks = array(
+        'woocommerce_email_footer_text'     => 20,
+        'woocommerce_email_from_address'    => 20,
+        'woocommerce_email_from_name'       => 20,
+        'woocommerce_attribute_taxonomies'  => 20,
+        'woocommerce_variation_option_name' => 20,
+    );
+
     $email_ids = array(
         'backorder'                         => 20,
         'cancelled_order'                   => 20,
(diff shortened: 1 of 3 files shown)


modules/woo-commerce/qwc-admin.php
```diff
@@ -26,27 +34,17 @@ function qtranxf_wc_add_filters_admin() {
...
+        $admin_hooks[ 'woocommerce_email_content_' . $name ]   = $priority;
     }
 
-    add_filter( 'woocommerce_attribute_taxonomies', 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
-    add_filter( 'woocommerce_variation_option_name', 'qtranxf_useCurrentLanguageIfNotFoundUseDefaultLanguage' );
+    qtranxf_add_filters( [ 'text' => $admin_hooks ] );
 }
 
 qtranxf_wc_add_filters_admin();
+global $wp_filter;
```
herrvigg, line 47: This doesn't seem to be used.

modules/woo-commerce/woo-commerce.php
```diff
@@ -42,3 +42,63 @@ function qtranxf_wc_detect_language( $url_info ) {
...
+     * In that case qtranxf_wc_add_filters_admin() can be called conditionally, following qtranxf_remove_filters call can be removed.
+     * Otherwise all filters affecting webhooks added in qtranxf_wc_add_filters_admin() must be removed here.
+     */
+    qtranxf_remove_filters( [
+        'text' => [
+            'woocommerce_attribute_taxonomies'  => 20,
+            'woocommerce_variation_option_name' => 20,
+        ]
+    ] );
+
```
herrvigg, line 82: Great you came up with this fix. 
However this webhook is a bit difficult to maintain as it's removing many filters set earlier so we have to be sure we don't miss any. Would be better to detect the webhook case earlier, but I guess this `woocommerce_deliver_webhook_async` action is fired too late.

modules/woo-commerce/qwc-admin.php
spleen1981: forgotten there... https://github.com/qtranslate/qtranslate-xt/pull/1194/commits/6fe1816eefd0847ec0ec174755df87574143495b

modules/woo-commerce/woo-commerce.php
spleen1981: > this webhook is a bit difficult to maintain

Definitely, but I did not find other ways, since as you guess the specific filter is coming too late to have a cleaner approach.
However we should take care mostly of the filters added in `qtranxf_wc_add_filters_admin()`, as `qtranxf_wc_add_filters_front()` is already prevented to be called if `wp_doing_cron()` is `true`.

PR #1209, opened 2022-07-28
QTS: replace slugs metabox only if existing

### LINKED ISSUE #1207: PHP error and irrelevant Slugs metabox in custom admin pages
Hello,

The Slugs module causes a PHP error on any admin pages using metaboxes if they are not a WP_Post:
[qtranslate-xt\modules\slugs\admin\slugs-admin.php line 130](https://github.com/qtranslate/qtranslate-xt/blob/3.12.0/modules/slugs/admin/slugs-admin.php#L130)

The reason is that the Slugs module assumes that the passed variable $post is a WP_Post object.
Of course, a page displaying metaboxes isn't necessarily a WP_Post.

A workaround would be to exit the qts_draw_meta_box() function early if the passed variable isn't a WP_Post.
For example, by adding at line 124:
`if( ! ( is_object( $post ) && ! empty( $post->ID ) ) ) { return; } `

A better solution would be to not display the "Slugs per language" metabox at all on irrelevant pages (that is, on pages that are not WP_Post).

spleen1981:
fixes #1207

### DIFF
--- a/modules/slugs/admin/slugs-admin.php
+++ b/modules/slugs/admin/slugs-admin.php
@@ -115,7 +115,7 @@ function qtranxf_slugs_add_slug_meta_box() {
     global $wp_meta_boxes;
 
     //Replace slugs metabox only if existing and not already removed
-    if ( ! empty($wp_meta_boxes[ get_current_screen()->id ][ 'normal' ][ 'core' ][ 'slugdiv' ] ) ) {
+    if ( ! empty( $wp_meta_boxes[ get_current_screen()->id ]['normal']['core']['slugdiv'] ) ) {
         remove_meta_box( 'slugdiv', null, 'normal' );
         add_meta_box( 'qts_sectionid', __( 'Slugs per language', 'qtranslate' ), 'qtranxf_slugs_draw_meta_box', null, 'side', 'high' );
     }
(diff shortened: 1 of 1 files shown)


modules/slugs/admin/slugs-admin.php
```diff
@@ -112,19 +112,13 @@ function qtranxf_slugs_deactivate() {
...
-    ];
-    if ( isset( $post_type ) && in_array( $post_type, $not_applicable_types ) ) {
-        return;
-    }
+    global $wp_meta_boxes;
 
-    remove_meta_box( 'slugdiv', null, 'normal' );
-    add_meta_box( 'qts_sectionid', __( 'Slugs per language', 'qtranslate' ), 'qtranxf_slugs_draw_meta_box', null, 'side', 'high' );
+    //Replace slugs metabox only if existing and not already removed
+    if ( ! empty($wp_meta_boxes[ get_current_screen()->id ][ 'normal' ][ 'core' ][ 'slugdiv' ] ) ) {
```
herrvigg, line 118: I can't test at this moment. Is this robust enough if one key is missing? I'm afraid `empty` will raise an exception if one key is missing - unlike `unset`.

spleen1981: Based on php.net,

> empty() is essentially the concise equivalent to !isset($var) || $var == false

which is what we need here, because that key may not be existing or may be `false` if already removed, so we should be covered.

PR #1211, opened 2022-07-28
Add function to check if a request is a GraphQL call.

EarthlingDavey:
Specifically for wp-graphql plugin: https://github.com/wp-graphql/wp-graphql - 3.3k Github stars and 20k active installs.

Add function to check if a request is a GraphQL call, and, if GraphQL call then don't do redirect or set cookies.

Without this change, WPGraphQL does not function correctly. It's default endpoint of /graphql is redirected to /xx/graphql

### DIFF
--- a/qtranslate_core.php
+++ b/qtranslate_core.php
@@ -215,8 +215,8 @@ function qtranxf_detect_language( &$url_info ) {
 
     $url_info['language'] = $lang;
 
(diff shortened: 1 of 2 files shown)


herrvigg (approved):
Looks good!

EarthlingDavey:
Thanks bud. I appreciate your's and the other contributor's work 👊👊✌️

Issue #1206, opened 2022-07-25
WordPress database error You can't specify target table 'wp_postmeta' / 'wp_termmeta' for update in FROM clause


yoancutillas:
Hello, I am getting an error right after migrating QTS Slugs:

```
[25-Jul-2022 12:09:57 UTC] WordPress database error You can't specify target table 'wp_postmeta' for update in FROM clause for query DELETE FROM wp_postmeta WHERE meta_key like 'qtranslate_slug_%' AND post_id in (SELECT DISTINCT(post_id) FROM wp_postmeta WHERE meta_key LIKE '_qts_slug_%') made by require_once('wp-admin/admin.php'), do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, qtranxf_admin_init, qtranxf_edit_config, qtranxf_executeOnUpdate, qtranxf_slugs_migrate_qts_data, qtranxf_slugs_migrate_qts_meta, {closure}
[25-Jul-2022 12:09:58 UTC] WordPress database error You can't specify target table 'wp_termmeta' for update in FROM clause for query DELETE FROM wp_termmeta WHERE meta_key like 'qtranslate_slug_%' AND term_id in (SELECT DISTINCT(term_id) FROM wp_termmeta WHERE meta_key LIKE '_qts_slug_%') made by require_once('wp-admin/admin.php'), do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, qtranxf_admin_init, qtranxf_edit_config, qtranxf_executeOnUpdate, qtranxf_slugs_migrate_qts_data, qtranxf_slugs_migrate_qts_meta, {closure}
```

The reason is probably that: https://stackoverflow.com/a/4562797

> you cannot modify the same table from a subquery within the same query. However, you can either SELECT then DELETE in separate queries, or nest another subquery and alias the inner subquery result (looks rather hacky, though):
> 
> ```
> DELETE FROM posts WHERE id IN (
>     SELECT * FROM (
>         SELECT id FROM posts GROUP BY id HAVING ( COUNT(id) > 1 )
>     ) AS p
> )
> ```

Indeed it is working properly if you replace [slugs-migrate-qts.php line 73](https://github.com/qtranslate/qtranslate-xt/blob/master/modules/slugs/admin/slugs-migrate-qts.php#L73)
[...]

PR #1216, opened 2022-08-31
Fix MySQL error #1206

spleen1981:
Fix MySQL error - You can't specify target table 'wp_postmeta' / 'wp_termmeta' for update in FROM clause
#1206


herrvigg (approved):
I still don't understand why I can only reproduce it with phpMyAdmin but not with the PHP code, but it's safe to merge in any case.

PR #1217, opened 2022-09-04
Escape `_` in SQL LIKE queries for Slugs import

herrvigg:
The previous code worked but `_` is a wildcard for any character in SQL LIKE.
Escape it to make the query more reliable in case the prefix values are changed (for developers).
Ignore other wildcards.


herrvigg:
Not directly related to #1206 but used in the same context.

Issue #1085, opened 2021-11-28
Is the plugin ready for new PHP 8.1?


SeoDeveloper01:
I'm going to upgrade PHP on my server to 8.1, so I want to make sure everything goes smoothly.

herrvigg:
Usually it's fine with minor upgrades (though they call it major).
But if you do experiments in production but it's at your own risk. Better try on dev sites first, 

First look at the depreciations, should be fine:
https://www.php.net/releases/8.1/en.php#deprecations_and_bc_breaks

But here I found something with `strftime` they didn't report on the official page:
https://php.watch/versions/8.1/strftime-gmstrftime-deprecated

There is an advanced option to use `strftime` in QTX but it's not the default. So for those using this you'll be forced to switch. We'll have to remove that option.

alphp:
Deprecated: preg_split(): Passing null to parameter #3 ($limit) of type int is deprecated in wp-content\plugins\qtranslate-xt-3.11.4\qtranslate_core.php on line 832:
Fixed:
```php
$vals = preg_split( '/[\s,]+/', strtolower( $ignore_file_types ), -1, PREG_SPLIT_NO_EMPTY );
```

For `strftime` you can use this gist: https://gist.github.com/bohwaz/42fc223031e2b2dd2585aab159a20f30

herrvigg:
More info about PHP 8.1
https://make.wordpress.org/core/2022/01/10/wordpress-5-9-and-php-8-0-8-1/

herrvigg:
I don't recommend migrating to PHP 8.1 yet.

The fix for `strftime` might take more time. It's used in more places than expected. But I don't think we want to rewrite that function, we should make it more specific to every usage.

Also, there are so many deprecated notices triggered in WP core, even in 5.9. They say it's ["just notices in debug mode, no impact on production, no need to panic"](https://core.trac.wordpress.org/ticket/54672). But in practice it's very annoying when for developers. It seems they had a huge list of PHP 8.1 deprecations such as [Requests](https://core.trac.wordpress.org/ticket/53635) to be fixed but only a few of them made their way in WP 5.9.
For sure there are also many plugins and themes that need to be upgraded.

alphp:
I'm doing the first impact assessments in test environments.
These tests allow me to detect incompatibilities with applications from other servers, such as databases.
It is in this scenario where I have tested the compatibility of WP with different plugins and themes that I use regularly.

herrvigg:
[...]

PR #1228, opened 2022-09-25
Replace `strftime` with `IntlDateFormatter` for PHP8.1

herrvigg:
`strftime` is deprecated in PHP8.1 and should not be used anymore. Add a locale-formatted strftime using `IntlDateFormatter`. This patch is a temporary workaround for immediate PHP8.1 support.

The `strftime` format should be abandonned completely in QTX so it becomes legacy from now. Mark all strftime methods as deprecated for PHPDoc but not for the PHP code yet, until a migration is done.

Attention: QTX uses an extended version of strftime to support extra format tags such as `%q` that are non-standard in strftime.

### DIFF
--- a/qtranslate_date_time.php
+++ b/qtranslate_date_time.php
@@ -1,11 +1,193 @@
 <?php
 
 /**
+ * Locale-formatted strftime using \IntlDateFormatter (PHP 8.1 compatible)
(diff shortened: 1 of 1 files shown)


herrvigg:
Should solve #1085 as a temporary workaround.

PR #1230, opened 2022-10-04
fix same translation for all langs not overwriting previous translations

### LINKED ISSUE #1215: Taxonomy term title doesn't get saved properly when the translated title is the same in multiple languages
Hi,

I've searched all over the internet and didn't find any thread related to this specific issue. 

So lets say I have French and German as languages and I have a taxonomy term with the title "Test" in french and "TestDE" in german.

Whenever I try to change the german title back to "Test" it doesn't get saved properly and it always keeps the old title "TestDE".
This issue only occurs when I pick exactly the same title as in the other language. For example if I try to change it to "TestDE2" it works.

One workaround is to use [:fr]Test[:de]Test[:] but it isn't always easy for the customer to use.

I'm also using the latest version of qtranslate-xt.

I hope anyone could help me here, many thanks in advance.

spleen1981:
fixes #1215

### DIFF
--- a/admin/qtx_admin.php
+++ b/admin/qtx_admin.php
@@ -47,11 +47,6 @@ function qtranxf_collect_translations( &$qfields, &$request, $edit_lang ) {
         if ( ! qtranxf_isMultilingual( $request ) ) {
             // convert to ML value
             $qfields[ $edit_lang ] = $request;
-            //Handles same translation for all languages to overwrite properly previous translations (change ignored otherwise)
-            if ( ! is_null( qtranxf_allthesame( $qfields ) ) ) {
-                array_walk( $qfields, function( &$value ){ $value=""; } );
-                $qfields[qtranxf_getLanguageDefault()] = $request;
-            }
             $request               = qtranxf_collect_translations_deep( $qfields, $sep );
         } else {
             // raw mode, or user mistakenly put ML value into an LSB-controlled field
--- a/admin/qtx_admin.php
+++ b/admin/qtx_admin.php
@@ -47,6 +47,11 @@ function qtranxf_collect_translations( &$qfields, &$request, $edit_lang ) {
         if ( ! qtranxf_isMultilingual( $request ) ) {
             // convert to ML value
             $qfields[ $edit_lang ] = $request;
+            //Handles same translation for all languages to overwrite properly previous translations (change ignored otherwise)
+            if ( ! is_null( qtranxf_allthesame( $qfields ) ) ) {
+                array_walk( $qfields, function( &$value ){ $value=""; } );
+                $qfields[qtranxf_getLanguageDefault()] = $request;
+            }
             $request               = qtranxf_collect_translations_deep( $qfields, $sep );
         } else {
             // raw mode, or user mistakenly put ML value into an LSB-controlled field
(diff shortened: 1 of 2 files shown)


Kevin-JP (approved):

admin/qtx_admin.php
```diff
@@ -47,6 +47,11 @@ function qtranxf_collect_translations( &$qfields, &$request, $edit_lang ) {
         if ( ! qtranxf_isMultilingual( $request ) ) {
             // convert to ML value
             $qfields[ $edit_lang ] = $request;
+            //Handles same translation for all languages to overwrite properly previous translations (change ignored otherwise)
+            if ( ! is_null( qtranxf_allthesame( $qfields ) ) ) {
+                array_walk( $qfields, function( &$value ){ $value=""; } );
+                $qfields[qtranxf_getLanguageDefault()] = $request;
+            }
```
herrvigg, line 54: When all the languages are the same we don't want to store ML tags, it should be stored as a simple string.
This works fine for post titles, so the bug is specific to terms. That part of code should not change.

herrvigg (changes requested):

admin/qtx_admin.php
spleen1981: agree, reverted previous commit and added term names specific fix in https://github.com/qtranslate/qtranslate-xt/pull/1230/commits/2a1486fd4dcb25627795730509c2944887b55a66

admin/qtx_admin_taxonomy.php
```diff
@@ -166,6 +166,9 @@ function qtranxf_term_sanitize_name_unslashed( $term, $taxonomy = null ) {
  */
 function qtranxf_term_sanitize_name_db( $term, $taxonomy = null ) {
     global $q_config;
+
+    if ( ! qtranxf_isMultilingual( $term ) ) return $term;
+
```
herrvigg, line 171: Yet it's unclear what to do with the slashes.
Even more surprising, if the term is not multilingual, the value returned by `qtranxf_term_sanitize_name_unslashed` is already a simple string (!), no ML tags. The reason is because it takes the default language anyway, even if the term is ML.
But... this change makes it work for some _other reason_, which is very convoluted... So, I don't understand yet how this works.
If we could simplify and remove all this mess of sanitization it would be nice. But there must (?) have been some reason for doing all of this.

spleen1981: >  the value returned by qtranxf_term_sanitize_name_unslashed is already a simple string 

Point is not the value returned to the filter, but the value set for `$q_config['terms_sanitized'][ $term ]` in that function, which is taken from the previous translations from options and used later in `qtranxf_term_set_translation` to be written again to options.
Hence for this scenario if term name is not multilingual call to `qtranxf_term_sanitize_name_unslashed` is avoided and `$q_config['terms_sanitized'][ $term ]` is not set (so it is removed from translations in options later).

Possible explanation for current behaviour of `qtranxf_term_sanitize_name_unslashed` (which may be needed in other scenarios not affected by this commit) here: https://github.com/qtranslate/qtranslate-xt/issues/1215#issuecomment-1279713113

herrvigg: Yes, got it. I was unsure if the calls for escaping with slashes were needed. But I tried with strings containing single quotes, double quotes and it works for all the cases (different values or all the same for each language).
Seems good now.

Issue #1234, opened 2022-10-16
Deprecate and migrate `strftime` formats for PHP8.1+


herrvigg:
The `strftime` function is deprecated with PHP8.1.
We have now a substitution called `qxtranxf_intl_strftime` (based on `IntlDateFormatter` working with PHP8.1) - see #1228.
This change is mostly transparent for the users, allowing to use in principle the current [strftime ](https://www.php.net/manual/en/function.strftime.php) or [date](https://www.php.net/manual/en/datetime.format.php) tags as before (default or by language). There is an advanced option called "Date / Time Conversion" to select the format. However, there are many caveats.

(A) The "strftime format" used in qTranslate has been extended to mimic some "date" formats.
For example: `%q` mimics date `S` for english day ordinal number (1st, 2nd, ...). That doesn't exist in strftime.
A full list can be found here: https://github.com/qtranslate/qtranslate-xt/issues/1085#issuecomment-1279944230. Most of these "extended tags" fill some empty gaps so it's acceptable, but a few ones are overriding existing official strftime tags. That's not good and those won't be supported anymore in their qTranslate definition. They look quite specific so it should not be a big problem.
```
date => strftime
'z' => '%F', // z: The day of the year (starting from 0) -- %F: Same as "%Y-%m-%d
'P' => '%s', // P: Difference to Greenwich time (GMT) with colon between hours and minutes -- %s: unix timestamp
'L' => '%k', // L: leap year -- %k: Hour in 24-hour format, single digit
```

(B) The "date" format is not handled natively by qTranslate.
It is converted to this "extended strformat" and then goes through a strftime call... But there are also some calls to [date](https://www.php.net/manual/en/function.date.php) to deal with the "extended format"... so it's a terrible mess. Moreover, the basic [date function](https://www.php.net/manual/en/function.date.php) does not care about locales so there are surely some limitations.

So, I suggest to abandon the strftime formats completely because it's such a mess to maintain. Here's the plan:
1. support the [date format](https://www.php.net/manual/en/datetime.format.php) natively, but we need to support it with the locales through `IntlDateFormatter`. It's different from the new strftime implementation, this doesn't exist yet.
2. deprecate the strftime values and **migrate** the strftime values to date. We need point 1 in place before doing this.
3. abandon strftime completely and remove all the legacy code

herrvigg:
It's even more complicated than expected. `IntlDateFormatter` is not following the date format but it has its own...
[...]

PR #1238, opened 2022-10-23
Refactor date-time conversions using `qtranxf_intl_strftime`

herrvigg:
The `strftime` function is deprecated in PHP8.1.
Generalize usage of `qtranxf_intl_strftime` for any PHP version. The old implementation is condemned so no point to maintain it. Eventually the strftime may disappear but this requires more work.

Abandon support of ML date-time formats.
The use case is very unclear and looks like very old legacy code. Simplify the code by assuming the format is only for the current lang.

Bump minimal PHP version to 5.5.
Required by `\DateTimeInterface`. Update composer with `ext-int` for `\IntlDateFormatter`.

Deprecations -> new versions
- `qtranxf_strftime` -> `qtranxf_intl_strftime`
- `qtranxf_convertFormat` -> `qtranxf_convert_to_strftime_format_using_config`. 
- `qtranxf_convertTimeFormat` -> `qtranxf_get_language_config_date_time`.
- `qtranxf_convertDateFormat` -> `qtranxf_get_language_config_date_time`.
-  `qtranxf_convertDateFormatToStrftimeFormat` -> `qtranxf_convert_date_format_to_strftime_format`.
- deprecate `$before` and `$after` arguments in `qtranxf_format_date` and `qtranxf_format_time`

### DIFF
--- a/admin/qtx_activation_hook.php
+++ b/admin/qtx_activation_hook.php
@@ -755,10 +755,10 @@ function qtranxf_clear_debug_log() {
 
 function qtranxf_activation_hook() {
     qtranxf_clear_debug_log();
-    if ( version_compare( PHP_VERSION, '5.4' ) < 0 ) {
+    if ( version_compare( PHP_VERSION, '5.5' ) < 0 ) {
         // Deactivate ourself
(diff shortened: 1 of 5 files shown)


herrvigg:
Related to #1085 #1228 #1234.

PR #1311, opened 2023-04-05
Moved deprecated functions to a separate file.

homeandriy:
These functions do not carry the logical load of the code in the files.

In the future, these functions will have to be removed


herrvigg:
Yes, better to isolate them. The only problem is if some developers include some PHP files specifically but that's a border case. 
I reformatted tabs to white spaces and move the main `require_once` to `init.php` (that should been seen as the main loader, I may rename it).

homeandriy:
Thank

I am preparing the next query based on the flags. It will be possible to choose an additional display of the flag in svg

Issue #32, opened 2018-06-21
Replaced the custom "xsplit" function with the standard JavaScript split function


herrvigg:
<a href="https://github.com/mweimerskirch"><img src="https://avatars3.githubusercontent.com/u/362092?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mweimerskirch](https://github.com/mweimerskirch)**
_Monday Feb 23, 2015 at 23:14 GMT_
_Originally opened as https://github.com/qTranslate-Team/qtranslate-x/pull/32_

----

I'm still trying to make the code more readable and easier to test and extend. IMHO the "xsplit" function can be removed. The function was necessary for qtranslate to be compatible with older browsers. Newer browser that implement ECMAScript 5 (this includes IE 9 in "standards mode") do not need this any longer. This would make the code a bit more readable, but IE 8 on the other hand would no longer be compatible. However, as WordPress displays a big warning when you use the admin area with IE 8 (or other old browsers), this should not be a problem.

----

_**[mweimerskirch](https://github.com/mweimerskirch)** included the following code: https://github.com/qTranslate-Team/qtranslate-x/pull/32/commits_

herrvigg:
<a href="https://github.com/mweimerskirch"><img src="https://avatars3.githubusercontent.com/u/362092?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mweimerskirch](https://github.com/mweimerskirch)**
_Wednesday Feb 25, 2015 at 10:46 GMT_

----

I tested the change in a lot of browsers. The only ones that failed were IE <= 8 (where WordPress displays a big warning message in the admin area so) and Konqueror (even in the most recent version). All other browsers I tested work fine. You can test for yourself if the "split" function is implemented correctly in a certain browser with this simple test page: http://michel.weimerskirch.net/xsplit-test.html (e.g. using browserstack.com).
If it's green, that means the split function captures the separator, if it's red, it didn't work and the "xsplit" workaround is still needed.
I added the same headers that WordPress uses, so IE works in the same "mode" than the WordPress admin interface.

IMHO this patch can be applied.

herrvigg:
The problem is not `split` but `@@split` related to the regex. Now that we have moved to webpack #990 the bundle uses `browserslist` which still supports IE up to version 11. It seems not even IE11 supports this feature:
https://caniuse.com/mdn-javascript_builtins_regexp_--split

[...]

PR #1313, opened 2023-04-07
Drop support for Internet Explorer

herrvigg:
WordPress abandoned the support of IE with WP5.8 in July 2021.
https://wordpress.org/news/2021/05/dropping-support-for-internet-explorer-11/

Though we don't require that version of WP, it's time to abandon that legacy. See also this very old ticket: #32.

1) Remove workaround for `RegExp:@@split`
See also https://caniuse.com/mdn-javascript_builtins_regexp_--split

2) Update our `browserslist` for JS to be synced with WordPress
https://github.com/WordPress/wordpress-develop/blob/trunk/package.json

3) Add more debug info about browser in the troubleshooting section.


herrvigg:
The change on the `browserlists` makes the JS code more compact and more modern e.g. with `const` (it was about time...).
I hope this won't break some old user configs but it is consistent with WP developer configuration that appeared in WP5.8.
Perhaps it's clearer if I make this as the required WP version though we don't need it strictly.

herrvigg:
I reverted the change on the new browserlist, it may break some old configurations beside IE.
We'll take it in a separate patch, maybe for next major release.

Issue #1080, opened 2021-11-15
Not loading on backend when using custom path for `wp-admin`


Splact:
Since I've a custom path for `wp-admin` for security reasons, the check [here](https://github.com/qtranslate/qtranslate-xt/blob/master/qtranslate_core.php#L890) is always failing when on admin pages. Seems like we are always `doing_front_end`, so `qtx_admin.php` is never loaded.

I fixed this changing the regex so that it matches my custom path, but would be nice to have here a wp filter so that we don't need to apply changes to the plugin codebase anymore.

herrvigg:
It reminds me of something, this part is handled pretty bad in WP. There should be a function to retrieve this, but there's only [get_admin_url](https://developer.wordpress.org/reference/functions/get_admin_url/) (URL, not path) and `wp_admin` is even hard-coded internally in WP... Do you get the correct path if run the `admin_url` filter?

Splact:
Yes I get the url with the right path, we could use this and rip off the protocol+domain part.

herrvigg:
Still it feels like a hack to use `get_admin_url`, it's not the same purpose and we should not assume about the http configuration.
I would also like to avoid adding a new filter just for this. Let's see if we can find a better solution.

herrvigg:
The main problem is that WordPress doesn't handle this properly, it's hard-coded in WP, it's pretty bad :-/

Issue #443, opened 2018-06-23
Incompatibility with WP Security Plugin due to hard-coded path


herrvigg:
<a href="https://github.com/aydancoskun"><img src="https://avatars2.githubusercontent.com/u/10948793?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [aydancoskun](https://github.com/aydancoskun)**
_Friday Oct 21, 2016 at 19:19 GMT_
_Originally opened as https://github.com/qTranslate-Team/qtranslate-x/issues/443_

----

I wrote the email below to the maintainer of the open source security plugin for wordpress which can be found at http://wp-hide.com. His response is below my mail. He is happy to assist to fix this.

**Subject:** WP Hide Security Enhancer and qtranslate-x compatibility
**Message:**
Dear Ladies and Gentleman,
First of all - VERY Nice work!! Just about priceless actually this plugin.
However I ran into an issue with qtranslate-x.
Basically what happens is qtranslate-x is not showing up / loading on the admin backend.
The site is still being translated and all is fine but the backend stopped working.
This is ONLY when I use the custom login / admin URL feature. All other options of wp-hide can be turned on everything works as expected.
The plugin is here:
https://github.com/qTranslate-Team/qtranslate-x
I looked into it and the problem is in this file:
https://github.com/qTranslate-Team/qtranslate-x/blob/master/qtranslate_core.php
Line 166:
`if(strpos($http_referer,'/wp-admin')!==FALSE){
`
Line 771:
`if(preg_match('#^/(wp-.*\.php|wp-login/|wp-admin/|xmlrpc.php|robots.txt|oauth/)#', $path)){
`
When I replace wp-admin with my custom wp-admin path it works but obviously this breaks upgradability.
I noticed that they hardcoded xmlrpc.php as well and that is probably not a good idea either.
I tried to fix this with using get_option calls and get_admin_url to replace their hardcoded code but had no success.
This is why I am calling on you.
I am a coder but wp is new to me. So I am calling on your expertise. If what they are doing is incorrect because they are hardcoding something, would you please be so kind to send a pull request to them or let me know what to do and I gladly will. They seem like a very responsive team.
If the problem is with your plugin, could I ask you to fix it please. I noticed that you have several "conflict" files in your code already to work around incompatibilities.
For the time being I have to stay with wp.admin path to keep the translators able to login and do their work. But I would love to change it.
I think the concept of your plugin is brilliant and very needed.
[...]

PR #1324, opened 2023-04-17
Fix custom URL for admin area and login page

poetter-sebastian:
If you use a plugin who hides the admin page on a custom url

Plugins like 
https://de.wordpress.org/plugins/wp-hide-security-enhancer/
https://de.wordpress.org/plugins/hide-my-wp/
https://de.wordpress.org/plugins/wps-hide-login/

According to #1080 #443


poetter-sebastian:
I created two new functions for backend and login name and used the str_replace instead of a regex.

src/url.php
```diff
@@ -320,6 +321,36 @@ function qtranxf_get_url_info( string $url ): array {
...
+ *
+ * @author Sebastian Poetter https://github.com/poetter-sebastian
+ * @link https://github.com/qtranslate/qtranslate-xt/pull/1324 repo pull request
+ *
+ * @return string WordPress backend name
+ */
+function qtranxf_get_admin_base():string
+{
+    $admin_base = str_replace( [site_url(), '/'], '', admin_url() );
+    return strlen($admin_base) == 0? 'wp-admin' : $admin_base;
```
spleen1981, line 336: - I'm not sure if there could be more complex cases where the custom base can contain slashes, it would be safer to change to `trim( str_replace( site_url(), '', admin_url() ), '/');`
-  No need to test here, we can just return the result of `trim` as in wp codebase `admin_url` is already built appending `wp-admin\` to `site_url`. Also the `wp-admin` could change upstream, though highly unlikely.

src/url.php
```diff
@@ -320,6 +321,36 @@ function qtranxf_get_url_info( string $url ): array {
...
+ *
+ * @author Sebastian Poetter https://github.com/poetter-sebastian
+ * @link https://github.com/qtranslate/qtranslate-xt/pull/1324 repo pull request
+ *
+ * @return string WordPress backend login name
+ */
+function qtranxf_get_login_base():string
+{
+    $login_base = str_replace( [site_url(), '/', '.php'], '', wp_login_url() );
+    return strlen($login_base) == 0? 'wp-login' : $login_base;
```
spleen1981, line 351: - Same as above, just use and return `trim`.
- There is something wrong with the original regex I think, as the `wp-login.php` case is already covered by the first block (`wp-.*\.php`). Anyways if you want to extend to the case of custom bases, there is no need to strip the `'.php'`, as what we want to get is actually `wp-login.php` rather than `wp-login` only. And `'.php'` could be part of the custom base also.

dev/qtx-test-test-urlBases.php
```diff
@@ -0,0 +1,83 @@
...
+        'wp-login',
+        'wp-login'
+    ];
+
+    qtranxf_check_url_replacement($urls, $exampleAdmin, $exampleAdminSolves);
+
+    qtranxf_check_url_replacement($urls, $exampleLogin, $exampleLoginSolves);
+}
+
+qtranxf_run_test_urlBases();
```
spleen1981, line 83: This is not needed, I would drop this file

spleen1981 (changes requested):

poetter-sebastian:
Thanks for the comments, I simplified the functions.

spleen1981 (approved):

herrvigg:
Looks good at first sight, I'll take a closer look soon.

src/url.php
```diff
@@ -320,6 +321,32 @@ function qtranxf_get_url_info( string $url ): array {
     return $urlinfo;
 }
 
+/**
+ * Returns the base admin url of the WordPress backend name e.g. wp-admin
+ *
+ * @author Sebastian Poetter https://github.com/poetter-sebastian
+ * @link https://github.com/qtranslate/qtranslate-xt/pull/1324 repo pull request
```
herrvigg, line 328: We don't want to track the authors by function, this is an open-source project.
Also the link is unnecessary.

src/url.php
```diff
@@ -320,6 +321,32 @@ function qtranxf_get_url_info( string $url ): array {
...
+function qtranxf_get_admin_base():string
+{
+    return trim( str_replace( site_url(), '', admin_url() ), '/');
+}
+
+/**
+ * Returns the base admin url of the WordPress backend login url e.g. wp-login.php
+ *
+ * @author Sebastian Poetter https://github.com/poetter-sebastian
+ * @link https://github.com/qtranslate/qtranslate-xt/pull/1324 repo pull request
```
herrvigg, line 341: We don't want to track the authors by function, this is an open-source project.
Also the link is unnecessary.

src/url.php
```diff
@@ -283,7 +283,8 @@ function qtranxf_language_neutral_path( string $path ): bool {
     if ( isset( $language_neutral_path_cache[ $path ] ) ) {
         return $language_neutral_path_cache[ $path ];
     }
-    if ( preg_match( '#^/(wp-.*\.php|wp-login/|wp-admin/|xmlrpc.php|robots.txt|oauth/)#', $path ) ) {
+
+    if ( preg_match( '#^/(wp-.*\.php|'.qtranxf_get_login_base().'/|'.qtranxf_get_admin_base().'/|xmlrpc.php|robots.txt|oauth/)#', $path ) ) {
```
herrvigg, line 287: We don't need to check these parts with a regex, it can be done more simply in two steps.

herrvigg (changes requested):
I'll try to rewrite this a bit differently.

src/url.php
herrvigg: The check conditions become also difficult to read.

src/modules/slugs/slugs.php
```diff
@@ -326,10 +326,10 @@ public function query_vars( $query_vars ) {
             }
 
             // If req_uri is empty or if it is a request for ourself, unset error.
-            if ( empty( $request ) || $req_uri == $self || strpos( $_SERVER['PHP_SELF'], 'wp-admin/' ) !== false ) {
+            if ( empty( $request ) || $req_uri == $self || strpos( $_SERVER['PHP_SELF'], qtranxf_get_admin_base().'/' ) !== false ) {
                 unset( $_GET['error'] );
                 unset( $error );
-                if ( isset( $perma_query_vars ) && strpos( $_SERVER['PHP_SELF'], 'wp-admin/' ) !== false ) {
+                if ( isset( $perma_query_vars ) && strpos( $_SERVER['PHP_SELF'], qtranxf_get_admin_base().'/' ) !== false ) {
```
herrvigg, line 332: @spleen1981 do you know why we use PHP_SELF here instead of REQUEST_URI? Is it fine if I rewrite it like this?
```
str_starts_with( $_SERVER['REQUEST_URI'], wp_admin_url() )
```

The whole `query_vars` function is a mess. I don't understand well the last check and I'm not sure the implementations of `qtranxf_get_admin_base()` work for all configurations (URL mode, ...).

src/url.php
herrvigg: Done. Please test the new version.

src/modules/slugs/slugs.php
spleen1981: > why we use PHP_SELF here instead of REQUEST_URI?

I would leave as it is, as `REQUEST_URI` is before URL rewrites (i.e. it's as per the original call URL) while `PHP_SELF` is after (i.e. it's as seen by PHP, not necessarily the original call URL).
Anyway still you need to compare with `qtranxf_get_admin_base()` instead of the entire `admin_url()`.

>  I'm not sure the implementations of qtranxf_get_admin_base() work for all configurations 

as `admin_url()` is built using `site_url()` + a suffix, I guess it will always work the opposite to retrieve that suffix.

herrvigg: All right, thanks for the answer. However I think the check is too "relaxed". We should not strip the slashes. A path should contain the slashes. By removing the slashes we could assume wrongly some other paths containing `wp-admin` partially (or the value that is customized).

herrvigg: One problem though is that PHP_SELF is a path to the local script, so it could contain backslashes on Windows.

herrvigg: Not entirely sure about the values we could get with PHP_SELF.
Imo we should also check the values at the beginning of the string, not at any position. We should not assume the website doesn't contain other sub-folders with that name. The previous checks were not good. I fixed this for the neutral paths but with HTTP_REFERER there's also the same question.

herrvigg: Mmm on Linux PHP_SELF doesn't start with a slash apparently unlike REQUEST_URI (and therefore the path given in neutral path but that should be fixed now).
Let's precise better these remain questions to finalize this PR:
- slashes vs backslashes (Windows?) in PHP_SELF (I don't know well which values this can take)
- maybe more precise checks for PHP_SELF and HTTP_REFERER to avoid assumptions

spleen1981: > We should not strip the slashes

agree, that's why we changed from `str_replace` to `trim` in `qtranxf_get_admin_base()`, in order to preserve possible inner slashes

> One problem though is that PHP_SELF is a path to the local script, so it could contain backslashes on Windows.

mmm I'm not sure, that would be applicable to `PATH` but not to `PHP_SELF`, which I think is obtained starting from the URL. At least with my windows XAMPP installation that is the situation.

herrvigg: > agree, that's why we changed from str_replace to trim in qtranxf_get_admin_base(), in order to preserve possible inner slashes

That's good but strictly speaking not sufficient, by "strip" I also include "trim". Slashes _might_ be removed, but in that case additional care should be taken for the position of the search. Looking for any stripped/trimmed string is bug-prone as in theory it can appear anywhere in the path, at any position. That's why I would rather keep the slashes but PHP_SELF is not exactly as REQUEST_URI regarding the format of the path.

poetter-sebastian:
Is there something I can do ?

src/modules/slugs/slugs.php
herrvigg: Let's consider the fixes go beyond scope of the patch to generalize to custom `wp-admin`. The patch should now be pretty conservative because it retrieves the same value for a default config. We'll handle possible fixes separately.

herrvigg:
I will merge it, but please test more once it's in master.
I spotted possible bugs but they come from the state before patch so let's separate the scope of changes. Beyond the custom `wp-admin` functionality coming here, this should also fix neutral path for `wp-login.php`. The previous version has a regex match against `wp-login/` (with trailing slash) but this looks wrong to me. The new check is a strict equality vs `wp_login_url()` which leads to `wp-login.php` by default.

herrvigg:
Released in 3.15.0.

Issue #1377, opened 2023-10-02
Fatal Error on the Woocommerce Shop Page when using product filters


charroma:
qtranslate-xt: Version 3.15.2 (latest)
Woocommerce: Version 8.1.1 (latest)
_PHP Fatal error: Uncaught TypeError: QTX_Module_Slugs::get_page_by_path(): Argument #1 ($page_path) must be of type string, null given, called in /qtranslate-xt/src/modules/slugs/slugs.php on line 438 and defined in /qtranslate-xt/src/modules/slugs/slugs.php:1039_

This happens on a Woocommerce Shop page (special product archive page) when the product filter is used and add parameters to the url.
ex: store/?orderby=popularity&paged=1
`$query['product']` doesn't exist in this specific case on line 437.

I see the code was changed last week, but not released yet. https://github.com/qtranslate/qtranslate-xt/pull/1363
The change will prevent the fatal error, but still leave a warning.

 _Warning: Undefined array key "product" in /qtranslate-xt/src/modules/slugs/slugs.php on line 441_
`441: $page_slug = $query['name'] ?? $query[ $query['post_type'] ];`

adding another null check would fix the problem for the future release
`441: $page_slug = $query['name'] ?? $query[ $query['post_type'] ] ?? null;`

spleen1981:
Agree, feel free to submit a PR for that.

charroma:
@spleen1981 I never submitted a PR before. I assumed it was easier to let the people already working on the project make the necessary change if they thought it was necessary.

spleen1981:
Of course it's easier. This project is community driven, anyone's contribution is welcome.

spleen1981:
this is closed, isn't it?

PR #1378, opened 2023-10-04
[SLUGS] fix warning on post_type missing key

charroma:
prevent warning on Woocommerce special product archive page when using the product filter


spleen1981 (approved):

spleen1981:
fixes #1377

PR #1404, opened 2024-04-02
Detailed instructions to migrate Slugs from legacy

mcamprecios:
Some extra instructions to migrate Slugs from the legacy plugin, after having some issues and getting help from one of the maintaners, @spleen1981.

Discussion can be followed here: https://github.com/qtranslate/qtranslate-xt/discussions/1401#discussioncomment-8950469


herrvigg:
Thanks! Looks good :)

PR #1415, opened 2024-08-19
fix missing menu_items with no translation

charroma:
Prevent menu items (taxonomy) being removed from menus when the term title is the same as the default language and the options "Show menu items in an alternative language when translation is not available for the selected language." is selected.

### DIFF
--- a/src/frontend.php
+++ b/src/frontend.php
@@ -106,6 +106,9 @@ function qtranxf_wp_get_nav_menu_items( $items, $menu, $args ) {
                             $term = wp_cache_get( $item->object_id, $item->object );
                             if ( $term ) {
                                 $item_title = $q_config['term_name'][ $term->name ][ $language ] ?? '';
+                                 if ($item_title == '' && $q_config['show_menu_alternative_language']) {
+                                	$item_title = $item->title;
+                                }
                                 if ( ! empty( $term->description ) ) {
                                     $item->description = $term->description;
                                 }


sviluppomania:
@herrvigg

herrvigg:
@charroma The change looks fine at first sight, but I don't understand this description:

> Prevent menu items (taxonomy) being removed from menus when the term title is the same as the default language 

Since the check is on
```
$item_title = $q_config['term_name'][ $term->name ][ $language ] ?? '';
```
This is not about the **default** language but the **current** language?

> when the term title is the same as ...

Where is that comparison done? The check seems to be on missing translation, but the case is for a non-multilingual item so I'm getting really confused.

charroma:
@herrvigg I added the same concept for the post title a little before in the file but didn't commit it.

There is no comparison done. It's just a check for non translated content and the admin content preference.

I noticed some pages and terms not showing in the menus because there was no translation saved in the current language.

The item title is empty because there is no translation saved.
It always happens when I only save the title of a post or taxonomy in one language because it's the same in all language.
So when the translation is not found, the code get the original title instead.
It works fine on title and content, but in the navigation, the item disapeared.

The line make the non translated item as an empty string:
`$item_title = $q_config['term_name'][ $term->name ][ $language ] ?? '';`

When it's empty and the user wants to show the alternate languague, the orinigal title is shown. Without this, non translated menu items are not shown.
`if ($item_title == '' && $q_config['show_menu_alternative_language']) {`

Issue #1414, opened 2024-08-17
Is qtranslate-XT still mantained/active? 


asbestomolesto:
**Describe the bug**
No updates since months ago

**To Reproduce**
1. try to search for an update
2. find nothing
3. Ask about the plugin state
4. Get zero response from owner

**Expected behavior**
Plugin need to be mantained at least for security reasons

**Additional context**
Is this another case of a translation plugin being abandoned? :(

bagaweb:
I really hope we'll get some news from @herrvigg 

There is also a discussion here: https://github.com/qtranslate/qtranslate-xt/discussions/1410

asbestomolesto:
> I really hope we'll get some news from @herrvigg
> 
> There is also a discussion here: #1410

Yep. I posted here as a bug so MAYBE we can get at least an answer :(

rafaelurben:
It's starting to look like qTranslate-XT will end up like [qTranslate](https://qian.qin.berlin/qtranslate/) and [qTranslate-X](https://github.com/qTranslate-Team/qtranslate-x/issues/579)... Abandoned. 😢 

Many things are already broken, especially due to the lack of full support for the Gutenberg editor.

The new WooCommerce product editor does not work at all with this plugin anymore (#1407), which is a deal breaker for me... I am currently thinking about switching to an alternative translation plugin.

bagaweb:
> Yep. I posted here as a bug so MAYBE we can get at least an answer :(

Yes let's hope to get an answer as soon as possible.

bagaweb:
> I am currently thinking about switching to an alternative translation plugin.

Do you know any similar plugins?

asbestomolesto:
I only know Polylang and we're evaluating it...

rafaelurben:
> Do you know any similar plugins?

I know that WPML and Polylang are popular alternatives, but I haven't analyzed them yet. I know of no similar plugin that stores translations inline and not in duplicate posts/products.

I have seen that WPML has a [migration tool](https://wpml.org/documentation/related-projects/qtranslate-importer/) to migrate away from qTranslate.

bagaweb:
> I know of no similar plugin that stores translations inline and not in duplicate posts/products.

Yes that's my biggest concern, finding a plugin similar to QTX. Personally I don't like the duplicate posts method.

rafaelurben:
> Yes that's my biggest concern, finding a plugin similar to QTX. Personally I don't like the duplicate posts method.

[...]

PR #1424, opened 2024-12-26
Issue-1422: Load the text domain using the "init" hook

phillip-hopper:
Fixes this WordPress warning: 

> PHP Notice:  Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>qtranslate</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later.

I removed the comment regarding the user not being authenticated because the user is authenticated when the "init" hook fires. 

https://developer.wordpress.org/reference/hooks/init/

### DIFF
--- a/qtranslate.php
+++ b/qtranslate.php
@@ -63,7 +63,7 @@ if ( ! defined( 'QTRANSLATE_FILE' ) ) {
 }
 
 require_once QTRANSLATE_DIR . '/src/init.php';
-add_action( 'plugins_loaded', 'qtranxf_init_language', 2 ); // User is not authenticated yet, high priority needed.
+add_action( 'init', 'qtranxf_init_language', 2 );
 
 if ( is_admin() ) {
     require_once QTRANSLATE_DIR . '/src/admin/activation_hook.php';


Komarovski (approved):

hirasso:
@herrvigg is qtranslate-xt still maintained? 🙏

rafaelurben:
@hirasso See https://github.com/qtranslate/qtranslate-xt/issues/1414#issuecomment-2297627536 for the most recent update on this matter.

vince6e74 (approved):

qtranslate.php
```diff
@@ -63,7 +63,7 @@
 }
 
 require_once QTRANSLATE_DIR . '/src/init.php';
-add_action( 'plugins_loaded', 'qtranxf_init_language', 2 ); // User is not authenticated yet, high priority needed.
+add_action( 'init', 'qtranxf_init_language', 2 );
```
herrvigg, line 66: The consequence of that change is MUCH bigger than just delaying the text domain.

herrvigg: This should do the trick in the init.php, please check it
```
add_action( 'init', 'qtranxf_load_plugin_textdomain' );
```

herrvigg:
Closed by https://github.com/qtranslate/qtranslate-xt/commit/5b71a8cc446a483d30a9032f5dd93a31b41a5b7e

qtranslate.php
Komarovski: Oh, I'm sorry, I have approved this review and didn't mentioned, that it is for "qtranslate.php" file, not for "init.php".

This fix:

`add_action( 'init', 'qtranxf_load_plugin_textdomain' );`

For "init.php" file works perfectly!

herrvigg: Nothing to be sorry about! Thanks for testing :)

PR #1436, opened 2025-04-26
Solution to small bug.

sviluppomania:
Changes the slugs home_url function to expect the path to null , fixing a small bug in the recall of user profile data in mobile apps.


herrvigg:
Released in 3.15.4.

PR #1491, opened 2026-06-15
Fix admin color fallback for PHP 8

Davesaggio:


Davesaggio:
## Summary

This fixes a PHP 8 fatal error in `qtranxf_get_user_admin_color()` when the current user does not have a valid admin color scheme available in `$_wp_admin_css_colors`.

Previously the code assumed this entry always existed and directly accessed:

`$_wp_admin_css_colors[ $user_admin_color ]->colors`

For non-admin users or unauthenticated/ajax contexts this could trigger a 500 error.

## Fix

- validate that the selected admin color exists
- validate that `->colors` exists and is an array
- fall back to `fresh` when the user color is missing/invalid
- return an empty array as a final safe fallback

## Notes

Verified with `php -l`.

PR #1494, opened 2026-06-15
Translate Yoast replacement variables before template processing

Komarovski:
Translate Yoast SEO replacement variable values before inserting them into multilingual metadata templates.

When a multilingual variable such as %%title%% was inserted inside a Raw ML template, it produced nested qTranslate language blocks that could not be parsed correctly.

Process string replacement values through the current-language fallback before Yoast expands its templates.

This prevents duplicated or incorrectly ordered text in SEO titles, descriptions, Open Graph metadata, and X/Twitter metadata while preserving multilingual templates.

### DIFF
--- a/src/modules/wp-seo/wp-seo-front.php
+++ b/src/modules/wp-seo/wp-seo-front.php
@@ -3,7 +3,38 @@ if ( ! defined( 'ABSPATH' ) ) {
     exit;
 }
 
+/**
+ * Translate Yoast replacement variables before they are inserted
+ * into multilingual SEO templates.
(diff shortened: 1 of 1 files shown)


src/modules/wp-seo/wp-seo-front.php
Komarovski: // TODO: Remove this filter. `wpseo_indexing_data` only modifies data
// passed to Yoast's indexing UI and does not prevent indexables from
// being created or saved. Returning false may also break the expected
// array structure.

